PR #1, opened 2021-07-09
Add ability to read yaml/yml files

jessfraz:
A bunch of the apis I use only come in yaml. Could convert to json outside of this tool but was fairly short.

Also cleaned up some clippy warnings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,6 +88,7 @@ dependencies = [
  "openapiv3",
  "serde",
  "serde_json",
+ "serde_yaml",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,4 +11,5 @@ anyhow = "1"
 getopts = "0.2"
 serde = { version = "1", features = [ "derive" ]}
 serde_json = "1"
+serde_yaml = "^0.8.17"
 openapiv3 = "0.4"
--- a/src/main.rs
+++ b/src/main.rs
@@ -272,7 +272,7 @@ impl ExtractJsonMediaType for openapiv3::Response {
     }
 
     fn is_binary(&self) -> Result<bool> {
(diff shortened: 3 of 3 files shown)


jclulow (commented):
Neat!  What are some of the API definitions you're trying?

jessfraz:
So I've tried a few, my biggesst need has been for a github API client that has all the endpoints and can easily update with new features, been working on it here: https://github.com/oxidecomputer/github-api-client/ will upstream the changes that make the most sense, lots of stuff for docs, enums, params, etc.

jessfraz:
One of my biggest hestiations is, for like the cio repo where I manually made clients I could be a bit opinionated about the API, and for something like github i was going to implement tags so the interface was more spread out than in one module, but for something like tailscale or shippo where the interface is smaller one module is good, so maybe when I contribute (my shit code that I will make better) there could be options for either or, basically I'm just experimenting with generating both and seeing the trade-offs.

jessfraz:
also the other thing that is super annoying me is no one uses the openapi spec configs the same way, like its nice there is a type for everything but if people don't use the spec right or add all the parameters like defaults for enums its super annoying, anyways just something to keep in mind for us that when we are generating our specs from dropshot we should be good citizens of the world

jessfraz:
the ecosystem is very inconsistent

jessfraz:
oh also ill flip that repo so it ultimately holds a bunch of third party apis so that way you can see the differences.

jessfraz:
example docs: https://rust.docs.corp.oxide.computer/github_api_client/

the docs look great :) but you can see the client (https://rust.docs.corp.oxide.computer/github_api_client/struct.Client.html) is a bit insane without tags putting things into categories

ahl:
Is this still needed? If so let me know and I'll resuscitate as there's been quite a bit of divergence.

PR #5, opened 2021-10-02
Add back whitespace

ahl:
@jclulow does this improve the whitespace situation?


jclulow:
It's a lot better, thank you!

I would trim the final blank line though, as git gets sad about it.

sad

PR #21, opened 2022-01-07
Allow consumers to specify additional derive macros for generated types

ahl:
This structures the parameters to the `generate_api!` macro (incompatibly) and allows for extra derive macros such as `JsonSchema`.

---8<--- old ---8<---

@smklein let me know if this works for you

I'm going to poke at two things
1. typify and progenitor optional user input for derive macros (it's tricky if we want to / need to be selective about the types to which these derives can be applied; maybe I'll make an unstable interface for this
2. progenitor emitting a dropshot server stub; I don't think it might be relatively easy given the IR we already have for operations

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -665,6 +665,7 @@ dependencies = [
  "progenitor-impl",
  "progenitor-macro",
  "reqwest",
+ "schemars",
  "serde",
  "serde_json",
  "uuid",
@@ -872,10 +873,12 @@ version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "271ac0c667b8229adf70f0f957697c96fafd7486ab7481e15dc5e45e3e6a4368"
 dependencies = [
+ "chrono",
  "dyn-clone",
  "schemars_derive",
  "serde",
  "serde_json",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


smklein:
First attempt at integrating (just adding a "Cargo.toml" patch for progenitor to use this branch):

```
error[E0432]: unresolved import `schemars`
 --> agent-client/src/lib.rs:5:1
  |
5 | generate_api!("../openapi/crucible-agent.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `schemars`
  |
  = note: this error originates in the macro `generate_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error: cannot determine resolution for the derive macro `JsonSchema`
 --> agent-client/src/lib.rs:5:1
  |
5 | generate_api!("../openapi/crucible-agent.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: import resolution is stuck, try simplifying macro imports
  = note: this error originates in the macro `generate_api` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I think the issue is that the "FOO-client" crates don't depend on schemars, but they would need to for this PR to work correctly.

I can go add that dependency (manually) and try once more.

smklein:
Adding the `schemars` dependencies, this seems to work! I'll upload branches shortly.

smklein:
The two linked PRs show the necessary changes. I would have used cargo patch to point to this branch, but that doesn't seem possible due to https://github.com/rust-lang/cargo/issues/5478. In the meantime, I just patched with local files to show what changes are incurred.

Issue #24, opened 2022-02-02
issues with untyped responses


jclulow:
I have an endpoint that streams file data where we don't really know the MIME type -- and it certainly isn't a structured JSON response:

```rust
#[endpoint {
    method = GET,
    path = "/0/jobs/{job}/outputs/{output}",
}]
pub(crate) async fn job_output_download(
    rqctx: Arc<RequestContext<Arc<Central>>>,
    path: TypedPath<JobsOutputsPath>,
) -> std::result::Result<Response<Body>, HttpError> {
...
    let mut res = Response::builder();
    res = res.header(CONTENT_TYPE, "application/octet-stream");

    let fr = c.file_response(&t.id, &o.id).await.or_500()?;
...
    res = res.header(CONTENT_LENGTH, fr.size);
    Ok(res.body(fr.body)?)
}
```

Critically, this is a generic `http::response::Response<hyper::body::Body>` that I have assembled myself rather than one of the Dropshot-provided typed responses.  In the past, Dropshot generated an OpenAPI definition like this:

```json
    "/0/jobs/{job}/outputs/{output}": {
      "get": {
        "operationId": "job_output_download",
        "parameters": [
          {
            "in": "path",
            "name": "job",
            "required": true,
            "schema": {
              "type": "string"
            },
            "style": "simple"
          },
          {
            "in": "path",
            "name": "output",
            "required": true,
            "schema": {
              "type": "string"
            },
            "style": "simple"
          }
        ],
        "responses": {}
      }
    },
```

Note that `responses` is empty.  I've updated my Dropshot version, though, and it now generates this instead:

```json
    "/0/jobs/{job}/outputs/{output}": {
      "get": {
        "operationId": "job_output_download",
        "parameters": [
          {
            "in": "path",
            "name": "job",
            "required": true,
            "schema": {
              "type": "string"
            },
            "style": "simple"
          },
          {
            "in": "path",
            "name": "output",
            "required": true,
            "schema": {
              "type": "string"
            },
            "style": "simple"
          }
        ],
        "responses": {
          "default": {
            "description": ""
          }
        }
      }
    },
```

This appears to confuse `progenitor`:

```
$ progenitor -i /ws/buildomat/cache/openapi.json -o /ws/buildomat/openapi -n buildomat-openapi -v 0.0.0
Error: op job_output_download: has response default
```

I don't really know what we _should_ be putting in the definition document for an untyped response, but I think in the meantime `progenitor` should probably handle this particular case.

[...]

PR #25, opened 2022-02-02
handle a different Dropshot expression of an untyped response

jclulow:


jclulow:
This is an attempt to fix (work around?) the issue I'm hitting with #24.

ahl:
I think that #26 will address this by handling the `default` field properly. It's currently ignoring the `description` field on responses; I could imagine using that if/when we handle multiple valid responses (otherwise that description field seems kind of pointless and it's somewhat vexing as to why OpenAPI requires it)

PR #36, opened 2022-03-16
update ByteStream type to include Pin

ahl:


jclulow:
Hey @ahl was there a reason you did `Send` but not `Sync` here?  I have added it myself in a local test and it solves some trouble I'm having in consuming the streams:

```diff
--- a/progenitor-client/src/lib.rs
+++ b/progenitor-client/src/lib.rs
@@ -14,7 +14,7 @@ use serde::de::DeserializeOwned;
 /// Represents a streaming, untyped byte stream for both success and error
 /// responses.
 pub type ByteStream =
-    Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>> + Send>>;
+    Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>> + Send + Sync>>;
```

I was otherwise seeing a lot of this sort of thing:

```
error[E0277]: `dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send` cannot be shared between threads safely
   --> client/src/main.rs:466:62
    |
466 |             let res = c.job_output_download(job, &o.id).await?;
    |                                                              ^ `dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send` cannot be shared between threads safely
    |
    = help: the trait `Sync` is not implemented for `dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send`
    = note: required because of the requirements on the impl of `Sync` for `Unique<dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send>`
    = note: required because it appears within the type `Box<dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send>`
    = note: required because it appears within the type `Pin<Box<dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send>>`
    = note: required because it appears within the type `ResponseValue<Pin<Box<dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send>>>`
    = note: required because it appears within the type `buildomat_openapi::Error<Pin<Box<dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send>>>`
    = note: required because of the requirements on the impl of `From<buildomat_openapi::Error<Pin<Box<dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send>>>>` for `anyhow::Error`
    = note: required because of the requirements on the impl of `FromResidual<Result<Infallible, buildomat_openapi::Error<Pin<Box<dyn futures::Stream<Item = Result<bytes::Bytes, reqwest::error::Error>> + std::marker::Send>>>>>` for `Result<(), anyhow::Error>`
```

ahl:
I only didn't add it because I hadn't needed it to consume the streams. Is there an easy test case that I could add that fails without `Sync`? Thanks!

jclulow:
I'm having trouble understanding what it wants, but I _suspect_ it is because I'm trying to use the `?` operator to convert the error type to an `anyhow::Error` automatically, and it might want `Sync` there.  I'll try and put something together!

ahl:
Don't worry about it if it's not easy to test. I just don't want it regressed in the future. And it's still uncomfortably easy to do so despite my best efforts regarding testing!

PR #42, opened 2022-04-26
Fix missing reqwest feature causing a compilation error

erikh:
Signed-off-by: Erik Hollensbe <git@hollensbe.org>


ahl (approved):
Thanks for the PR and apologies for the delayed response.

Looks good!

PR #75, opened 2022-05-26
pushing all path-wide parameters down to each method

rrichardson:
This solves the "parameters that apply to all path items"  issue.  
I'm not super familiar with the code, so I don't know if it works universally, but it works for my project  (Hashicorp Vault API)


ahl:
This looks like a good approach. Can you add some tests to validate this?

ahl:
I think this was covered by @262. Thank you for this submission and apologies for the long delay.

PR #93, opened 2022-06-22
openssl make optional

jessfraz:
this adds features for which reqwests to use. Let me know if you want me to extend your github action to test the various feature flags, these mirror those in third-party-api-clients

### DIFF
--- a/progenitor-client/Cargo.toml
+++ b/progenitor-client/Cargo.toml
@@ -10,6 +10,6 @@ description = "An OpenAPI client generator - client support"
 bytes = "1.1.0"
 futures-core = "0.3.21"
 percent-encoding = "2.1"
-reqwest = { version = "0.11", features = ["json", "stream"] }
+reqwest = { version = "0.11", default-features = false, features = ["json", "stream"] }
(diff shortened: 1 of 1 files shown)


ahl:
I'm not sure I follow how this works. Presumably this is for a client generated via the builder interface only since that's the only style of generation that includes an explicit dependency on `progenitor-client`. Is that right?

jessfraz:
oh I intended for both, want me to change it, I think it would just be then another nested feature flag

ahl:
I'm confused about how this works. My understanding of `cargo` is that it takes the union of all feature flags for a given dependency. So consumers would just use dependencies such as `pem` or `openssl` as needed. If the problem is that `reqwest` implies a TLS dependency by default, would it suffice to just set `default-features = false` for the `reqwest` dependency?

I don't doubt that there's a problem this solves, but I don't understand what it is.

jessfraz:
yeah I guess the easiest solution would be default-features false if progenitor-client does not need tls specifically

does it not?

ahl:
Doesn't seem to!

PR #107, opened 2022-07-05
Add application/x-www-form-urlencoded to extract json media type

karencfv:
Does what it says on the tin :)

On the [line](https://github.com/oxidecomputer/progenitor/blob/main/progenitor-impl/src/method.rs#L1865) above the ExtractJsonMediaType trait, it says "// TODO do I want/need this?" I'm guessing this might all be deleted in the future?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-/target
+target/
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -1876,9 +1876,13 @@ impl ExtractJsonMediaType for openapiv3::RequestBody {
 
         if let Some(mt) = self.content.get("application/json") {
             Ok(mt.clone())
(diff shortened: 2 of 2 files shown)


ahl (commented):
I think this will generate code that doesn't work. The generated client will still pass data as `
"application/json"` in all cases which isn't quite right. 

I was looking into how to properly fix this. It will involve a new kind of `OperationParameterType` that indicates the MIME type and the generated code will need to do the encoding manually via `serde_urlencoded` rather than using the `reqwest::RequestBuilder::json()` method as we do today.

I think the shortest path to getting this working may just be to delete these endpoints from the OpenAPI document as I do in this WIP: https://github.com/oxidecomputer/omicron/pull/1349

karencfv:
lol not as easy as I thought it was 😂 Thanks for your input!

PR #109, opened 2022-07-07
add support for application/x-www-form-urlencoded bodies

ahl:
@plotnick  and @karencfv you might be interested in this, but no need to review if not. If I merge before you get to this I'll still attend to any review comments you might leave.

### DIFF
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -27,7 +27,6 @@ pub(crate) struct OperationMethod {
     summary: Option<String>,
     description: Option<String>,
     params: Vec<OperationParameter>,
-    raw_body_param: bool,
     responses: Vec<OperationResponse>,
     dropshot_paginated: Option<DropshotPagination>,
 }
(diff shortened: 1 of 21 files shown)


alex:
First time today! :-)

progenitor-impl/src/method.rs
```diff
@@ -1655,6 +1675,100 @@ impl Generator {
...
+
+        let content_type = BodyContentType::from_str(content_str)?;
+
+        let typ = match content_type {
+            BodyContentType::OctetStream => {
+                // For an octet stream, we expect a simple, specific schema:
+                // "schema": {
+                //     "type": "string",
+                //     "format": "binary"
+                // }
```
karencfv, line 1710: Not sure if you meant to uncomment this code?

karencfv (approved):
Thanks for this! It'll be great to use the whole current nexus.json 🙇‍♀️

progenitor-impl/src/method.rs
ahl: No that's the JSON schema

plotnick (commented):
A little late to the party, but this looks good to me (though I admit to being unfamiliar with the progenitor code).

Issue #76, opened 2022-05-27
not yet implemented: expected one content entry, found 4


brainstorm:
Thanks for this library! I'm giving it a go for work purposes (https://github.com/brainstorm/ica-rs), unfortunately:

```
% cargo run
   Compiling ica-rust v0.1.0 (/Users/rvalls/dev/umccr/ica-rust)
error: proc macro panicked
 --> src/main.rs:4:5
  |
4 |     generate_api!(spec = "openapi/gds.json");
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: not yet implemented: expected one content entry, found 4
```

I'll keep digging, I just wanted to make sure whether you or somebody else was planning to support multiple content entries before I deep dive on it ;)

brainstorm:
So, the detected media types are the following:

```rust
% cargo run
   Compiling ica-rust v0.1.0 (/Users/rvalls/dev/umccr/ica-rust)
[/Users/rvalls/dev/umccr/progenitor/progenitor-impl/src/method.rs:1261] &self.content = {
    "application/json-patch+json": MediaType {
        schema: Some(
            Reference {
                reference: "#/components/schemas/VolumeFileListRequest",
            },
        ),
        example: None,
        examples: {},
        encoding: {},
        extensions: {},
    },
    "application/json": MediaType {
        schema: Some(
            Reference {
                reference: "#/components/schemas/VolumeFileListRequest",
            },
        ),
        example: None,
        examples: {},
        encoding: {},
        extensions: {},
    },
    "text/json": MediaType {
        schema: Some(
            Reference {
                reference: "#/components/schemas/VolumeFileListRequest",
            },
        ),
        example: None,
        examples: {},
        encoding: {},
        extensions: {},
    },
    "application/*+json": MediaType {
        schema: Some(
            Reference {
                reference: "#/components/schemas/VolumeFileListRequest",
            },
        ),
        example: None,
        examples: {},
        encoding: {},
        extensions: {},
    },
}
error: proc macro panicked
 --> src/main.rs:4:5
  |
4 |     generate_api!("./openapi/gds.json");
  |      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: not yet implemented: expected one content entry, found 4

error: could not compile `ica-rust` due to previous error
```

[...]

PR #111, opened 2022-07-08
improved template parsing

jclulow:
As mentioned in [a comment on #76](https://github.com/oxidecomputer/progenitor/issues/76#issuecomment-1178645832), our template handling is presently too strict: it requires that each parameter in the path spans an entire path component, with no additional strings.  We should be a bit more relaxed about this.


ahl (approved):
let's get that money

PR #119, opened 2022-07-12
Show how to use Client::new_with_client

karencfv:
Example documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-/target
+target/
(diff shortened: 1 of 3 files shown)


ahl (commented):
What do you think about putting this in the docs instead?
https://github.com/oxidecomputer/progenitor/tree/main/docs

karencfv:
Let me know if this is what you were after. Happy to change it if it's not :)

ahl:
Great! Thank you!

PR #122, opened 2022-07-15
oxidecomputer/progenitor#38 progenitor-produced Cargo.toml is a workspace root?

### LINKED ISSUE #38: progenitor-produced Cargo.toml is a workspace root?
It seems like #26 (I think?) introduced an empty `[workspace]` section to the generated `Cargo.toml`:

https://github.com/oxidecomputer/progenitor/blob/75348207d7405db3371a98b5348bef3fbd0ef79e/progenitor/src/main.rs#L93-L94

Is that intentional?  It confuses `cargo build` at least, when used to generate a crate that is already within a workspace:

```
$ cargo build
error: multiple workspace roots found in the same workspace:
  /ws/lux/keeper/openapi
  /ws/lux/keeper
```

jclulow:
seeks to fix #38

### DIFF
--- a/progenitor/src/main.rs
+++ b/progenitor/src/main.rs
@@ -128,8 +128,7 @@ fn main() -> Result<()> {
                 \n\
                 [dependencies]\n\
                 {}\n\
-                \n\
-                [workspace]\n",
+                \n",
                 name,
                 version,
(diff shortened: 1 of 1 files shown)


ahl:
Fine.

PR #123, opened 2022-07-15
ResponseValue could expose Content-Length

jclulow:
Add a `content_length()` method to `ResponseValue` which would return a numeric value iff `Content-Length` is present in the response and is a valid positive integer.

### DIFF
--- a/progenitor-client/src/progenitor_client.rs
+++ b/progenitor-client/src/progenitor_client.rs
@@ -118,6 +118,16 @@ impl<T> ResponseValue<T> {
         &self.headers
     }
 
+    /// Get the parsed value of the Content-Length header, if present and valid.
+    pub fn content_length(&self) -> Option<u64> {
+        self.headers
+            .get(reqwest::header::CONTENT_LENGTH)?
+            .to_str()
+            .ok()?
+            .parse::<u64>()
+            .ok()
+    }
+
     #[doc(hidden)]
     pub fn map<U: std::fmt::Debug, F, E>(
         self,


ahl:
Is this just something you found yourself doing often in consumers?

jclulow:
Yes, for untyped/binary responses I end up wanting the content length value a bunch.

progenitor-client/src/progenitor_client.rs
```diff
@@ -118,6 +118,14 @@ impl<T> ResponseValue<T> {
         &self.headers
     }
 
+    /// Get the parsed value of the Content-Length header, if present and valid.
+    pub fn content_length(&self) -> Option<u64> {
+        self.headers
+            .get(reqwest::header::CONTENT_LENGTH)
+            .and_then(|cl| cl.to_str().ok())
+            .and_then(|cl| cl.parse::<u64>().ok())
```
ahl, line 126: Is this the same as this?

```rust
        self.headers
            .get(reqwest::header::CONTENT_LENGTH)?
            .to_str()
            .ok()?
            .parse::<u64>()
            .ok()
 ```

I'm fine either way... just offering that up in case you find it to be more legible.

ahl (approved):

progenitor-client/src/progenitor_client.rs
jclulow: Ah, yeah, that's probably better.  I forget that they did `?` for `Option` provided you aren't doing any `Result` business as well.

PR #124, opened 2022-07-15
code generation fixes after 03e2cfad3c4d63

jclulow:
It looks like 03e2cfad3c4d637918b9e4a267f6a0ef426b0186 introduced a couple of small issues into the generated `Cargo.toml` when using the `progenitor` CLI:

* `serde_urlencoded` uses an unquoted version string, which the TOML parser really wants to believe is a float; this upsets `cargo`.
* `serde_json` is no longer optional based on the use of JSON types in the type space, as it is used unconditionally in the `form_urlencoded()` member of `RequestBuilderExt` for `RequestBuilder`.

### DIFF
--- a/progenitor-impl/src/lib.rs
+++ b/progenitor-impl/src/lib.rs
@@ -400,7 +400,7 @@ impl Generator {
             "percent-encoding = \"2.1\"",
             "reqwest = { version = \"0.11\", features = [\"json\", \"stream\"] }",
             "serde = { version = \"1.0\", features = [\"derive\"] }",
-            "serde_urlencoded = 0.7",
+            "serde_urlencoded = \"0.7\"",
         ];
         if self.type_space.uses_regress() {
             deps.push("regress = 0.4")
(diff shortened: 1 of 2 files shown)


progenitor-impl/src/lib.rs
```diff
@@ -400,7 +400,8 @@ impl Generator {
             "percent-encoding = \"2.1\"",
             "reqwest = { version = \"0.11\", features = [\"json\", \"stream\"] }",
             "serde = { version = \"1.0\", features = [\"derive\"] }",
-            "serde_urlencoded = 0.7",
+            "serde_json = \"1.0\"",
+            "serde_urlencoded = \"0.7\"",
```
ahl, line 404: form_urlencoded was supposed to use serde_urlencoded not serde_json

jclulow: https://github.com/oxidecomputer/progenitor/blob/0c7cace799bde609c87d1aeb9b093bc12d14f180/progenitor-client/src/progenitor_client.rs#L338

here?

ahl: Yes. I'm happy to sort that out if you'd prefer; it was an oversight on my part

ahl (approved):
let's get that money

PR #125, opened 2022-07-15
allow ResponseValue<ByteStream> to transfer ownership of interior Stream

jclulow:
There are a few places in buildomat where our dropshot servers are effectively acting as a proxy for large assets; e.g., the core API service stores objects in S3 and we get a byte stream back from the S3 client, and the frontend GitHub interface speaks with the core buildomat API service to get access to various assets.

It is perhaps surprisingly easy to arrange this, provided you can pass an owned stream of bytes to `hyper::Body::wrap_stream()`, allowing it to convert this into an appropriate `Body` for use with `hyper::Response::builder()` that we can then give to dropshot.  To do this, though, a few conditions need to be met that are not presently:

* `ResponseValue<ByteStream>` in the progenitor-derived client needs to be willing to give up its interior `Stream`
* That `Stream` needs to match the expected signature, which includes both `Sync` and `Send`

An example of roughly how this is used, in the context of a dropshot endpoint handler:

```rust
/*
 * Obtain a response from a progenitor-derived client, where the endpoint
 * is producing a large, untyped (binary) response:
 */
let response = buildomat_client.job_output_download(job_id, output_id).await?;

/*
 * Get the Content-Length and Content-Type that we want to pass on to our
 * own client:
 */
let clen = response.content_length().unwrap();
let ctype = guess_mime_type(filename);

/*
 * Generate a response with those custom headers, plus whatever data ends up
 * getting streamed from the backend service (without buffering it all in
 * memory):
 */
Ok(Some(hyper::Response::builder()
    .status(hyper::StatusCode::OK)
    .header(hyper::header::CONTENT_TYPE, ctype)
    .header(hyper::header::CONTENT_LENGTH, clen)
    .body(hyper::Body::wrap_stream(response.into_inner_stream()))?,
))
```

### DIFF
--- a/progenitor-client/src/progenitor_client.rs
+++ b/progenitor-client/src/progenitor_client.rs
@@ -16,11 +16,20 @@ use serde::{de::DeserializeOwned, Serialize};
 
 /// Represents an untyped byte stream for both success and error responses.
 pub struct ByteStream(
-    Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>> + Send>>,
+    Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>> + Send + Sync>>,
 );
 
+impl ByteStream {
+    pub fn into_inner(
+        self,
+    ) -> Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>> + Send + Sync>> {
+        self.0
+    }
+}
+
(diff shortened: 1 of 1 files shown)


ahl (commented):
ship it

PR #178, opened 2022-08-27
sync w/ typify and add a test

ahl:
@luqmana FYI


progenitor-impl/tests/output/test_default_params_positional.out
```diff
@@ -31,6 +33,10 @@ pub mod types {
         {
             T::try_from(V).unwrap()
         }
+
+        pub(super) fn body_with_defaults_something() -> Option<bool> {
```
ahl, line 37: this is where we used to have `super::Option<bool>`

PR #183, opened 2022-08-31
Generate a Client method for Dropshot websocket channels

lifning:
(rel: https://github.com/oxidecomputer/dropshot/pull/403 )
Generated methods return `ResponseValue<reqwest::Upgrade`, which may be passed to a websocket protocol implementation such as `tokio_tungstenite::WebSocketStream::from_raw_stream(rv.into_inner(), ...)` for the purpose of implementing against the raw websocket connection, but may later be extended as a generic to allow higher-level channel message definitions. (rel: https://github.com/oxidecomputer/dropshot/issues/429)

Per the changelog, consumers will need to depend on reqwest from git until such time upstream cuts a new crates.io release:
```
reqwest = {
    git = "https://github.com/seanmonstar/reqwest",
    rev = "6ceb23958c8b6d7f1d8ee093f0ad73184d133d40",
    features = ["json", "stream"]
}
```

### DIFF
--- a/progenitor-client/Cargo.toml
+++ b/progenitor-client/Cargo.toml
@@ -10,7 +10,7 @@ description = "An OpenAPI client generator - client support"
 bytes = "1.2.1"
 futures-core = "0.3.24"
 percent-encoding = "2.2"
-reqwest = { version = "0.11", default-features = false, features = ["json", "stream"] }
+reqwest = { version = "0.11.12", default-features = false, features = ["json", "stream"] }
 serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.7.1"
--- a/progenitor-impl/src/lib.rs
+++ b/progenitor-impl/src/lib.rs
@@ -26,6 +26,8 @@ pub enum Error {
     UnexpectedFormat(String),
     #[error("invalid operation path {0}")]
     InvalidPath(String),
+    #[error("invalid dropshot extension use: {0}")]
+    InvalidExtension(String),
     #[error("internal error {0}")]
     InternalError(String),
 }
@@ -36,6 +38,7 @@ pub struct Generator {
     type_space: TypeSpace,
     settings: GenerationSettings,
     uses_futures: bool,
+    uses_websockets: bool,
 }
 
 #[derive(Default, Clone)]
@@ -116,6 +119,7 @@ impl Default for Generator {
             ),
             settings: Default::default(),
             uses_futures: Default::default(),
+            uses_websockets: Default::default(),
         }
     }
 }
@@ -133,6 +137,7 @@ impl Generator {
             type_space: TypeSpace::new(&type_settings),
             settings: settings.clone(),
             uses_futures: false,
+            uses_websockets: false,
         }
     }
 
@@ -426,7 +431,7 @@ impl Generator {
             "bytes = \"1.1\"",
             "futures-core = \"0.3\"",
             "percent-encoding = \"2.1\"",
-            "reqwest = { version = \"0.11\", features = [\"json\", \"stream\"] }",
+            "reqwest = { version = \"0.11.12\", features = [\"json\", \"stream\"] }",
             "serde = { version = \"1.0\", features = [\"derive\"] }",
             "serde_urlencoded = \"0.7\"",
         ];
@@ -443,6 +448,10 @@ impl Generator {
         }
         if self.uses_futures {
             deps.push("futures = \"0.3\"")
+        }
+        if self.uses_websockets {
+            deps.push("base64 = \"0.13\"");
+            deps.push("rand = \"0.8\"");
         }
         if self.type_space.uses_serde_json() {
             deps.push("serde_json = \"1.0\"")
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -29,6 +29,7 @@ pub(crate) struct OperationMethod {
     params: Vec<OperationParameter>,
     responses: Vec<OperationResponse>,
     dropshot_paginated: Option<DropshotPagination>,
+    dropshot_websocket: bool,
 }
 
 enum HttpMethod {
@@ -189,6 +190,7 @@ impl OperationResponseStatus {
         matches!(
             self,
             OperationResponseStatus::Default
+                | OperationResponseStatus::Code(101)
                 | OperationResponseStatus::Code(200..=299)
                 | OperationResponseStatus::Range(2)
         )
@@ -225,6 +227,7 @@ enum OperationResponseType {
     Type(TypeId),
     None,
     Raw,
+    Upgrade,
 }
 
 impl Generator {
@@ -338,6 +341,12 @@ impl Generator {
             })
             .collect::<Result<Vec<_>>>()?;
 
+        let dropshot_websocket =
+            operation.extensions.get("x-dropshot-websocket").is_some();
+        if dropshot_websocket {
+            self.uses_websockets = true;
+        }
+
         if let Some(body_param) = self.get_body_param(operation, components)? {
             params.push(body_param);
         }
@@ -378,9 +387,10 @@ impl Generator {
                 let (status_code, response) = v?;
 
                 // We categorize responses as "typed" based on the
-                // "application/json" content type, "raw" if there's any other
-                // response content type (we don't investigate further), or
-                // "none" if there is no content.
+                // "application/json" content type, "upgrade" if it's a
+                // websocket channel without a meaningful content-type,
+                // "raw" if there's any other response content type (we don't
+                // investigate further), or "none" if there is no content.
                 // TODO if there are multiple response content types we could
                 // treat those like different response types and create an
                 // enum; the generated client method would check for the
@@ -407,6 +417,8 @@ impl Generator {
                     };
 
                     OperationResponseType::Type(typ)
+                } else if dropshot_websocket {
+                    OperationResponseType::Upgrade
                 } else if response.content.first().is_some() {
                     OperationResponseType::Raw
                 } else {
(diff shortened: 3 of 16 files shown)


progenitor-client/Cargo.toml
```diff
@@ -9,8 +9,16 @@ description = "An OpenAPI client generator - client support"
...
 reqwest = { version = "0.11", default-features = false, features = ["json", "stream"] }
 serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.7.1"
+
+# deps for websocket support
+base64 = "0.13"
+rand = "0.8"
+# reqwest wraps hyper, but does not expose a read/write socket for us to work with,
+# so we must make direct socket connections.
```
luqmana, line 23: My reqwest PR landed and added support for doing HTTP upgrades. It would mean a slightly different interface in that we could just do the upgrade in the generated code and return some type that impls `AsyncRead/AsyncWrite`. The user of the client lib could then use `tokio_tungstenite::WebSocketStream::from_raw_socket` or similar.

luqmana: That would also avoid requiring clients picking up new deps (although tbf, those are already in the dep chain even if not explicit).

lifning: ohh, even better!
i guess pending a proper release i'll just point it at the git repo

ahl:
It would be great to see a release that includes https://github.com/seanmonstar/reqwest/pull/1376 to move forward with this

progenitor-impl/src/method.rs
```diff
@@ -386,7 +393,9 @@ impl Generator {
                 // enum; the generated client method would check for the
                 // content type of the response just as it currently examines
                 // the status code.
-                let typ = if let Some(mt) =
+                let typ = if dropshot_websocket {
+                    OperationResponseType::Upgrade
```
ahl, line 397: Is the Content-type meaningless for the response from one of these endpoints?
https://stackoverflow.com/a/39105228/1371209

I assume it will never be `application/json` (is that true?)

Could we do
```rust
let typ = if let Some(mt) = response.content.get("application/json") {
    // ...
} else if !dropshot_websocket {
    OperationResponseType::Raw
} else {
    OperationResponseType::Upgrade
}
```

My preference (when it doesn't suck!) is to have the typical case first. And can we please update the comment above to cover the new case as well?

progenitor-client/Cargo.toml
```diff
@@ -10,7 +10,10 @@ description = "An OpenAPI client generator - client support"
...
 futures-core = "0.3.24"
 percent-encoding = "2.1"
-reqwest = { version = "0.11", default-features = false, features = ["json", "stream"] }
+reqwest = { git = "https://github.com/seanmonstar/reqwest", rev = "6ceb23958c8b6d7f1d8ee093f0ad73184d133d40", features = ["json", "stream"] }
+#reqwest = { version = "0.11", default-features = false, features = ["json", "stream"] }
 serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.7.1"
+base64 = "0.13"
+rand = "0.8"
```
ahl, line 19: I think these would need to be added here as well: https://github.com/oxidecomputer/progenitor/blob/main/progenitor-impl/src/lib.rs#L426

Is there a reasonable way to avoid these dependencies? "No" is a fine answer...

progenitor-impl/src/method.rs
```diff
@@ -842,6 +887,13 @@ impl Generator {
                             ))
                         }
                     }
+                    OperationResponseType::Upgrade => {
+                        quote! {
+                            Err(Error::ErrorResponse(
+                                ResponseValue::upgrade(response).await?
+                            ))
+                        }
+                    }
```
ahl, line 896: In what situation would we hit this?

sample_openapi/propolis-server.json
```diff
@@ -0,0 +1,850 @@
...
+        "operationId": "instance_serial",
+        "responses": {
+          "default": {
+            "description": "",
+            "content": {
+              "*/*": {
+                "schema": {}
+              }
+            }
+          }
```
ahl, line 161: In this case, what would the proper handling of a 4xx or 5xx response be? I realize that this is more germane for the dropshot code, but I'm thinking through how clients (direct or through progenitor) would handle it.

progenitor/tests/build_propolis.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright 2022 Oxide Computer Company
+
+// ensure that the websocket channel used for serial console compiles.
```
ahl, line 3: we might want to see the output as well along these lines:

https://github.com/oxidecomputer/progenitor/blob/main/progenitor-impl/tests/test_output.rs#L48

I know it's a little silly that we have two tests for each file, but one is for examining output and the other is just to make sure it compiles (with some spot checks).

ahl (commented):
very cool; it's a bummer that we'd need that git dependency, but I guess what can you do. If we do move forward with that, let's pause with merging this PR: I'd like to make a public release first to crates.io

Also it looks like some output needs updating to pass the tests.

lifning:
> very cool; it's a bummer that we'd need that git dependency, but I guess what can you do. If we do move forward with that, let's pause with merging this PR: I'd like to make a public release first to crates.io

thinking about it more, would it perhaps make sense for us to just re-export everything the generated code needs from our crate, and have the output import them that way so consumers don't have to garden their dependencies for updates like this in general? (like, git or not)

ahl:
> thinking about it more, would it perhaps make sense for us to just re-export everything the generated code needs from our crate, and have the output import them that way so consumers don't have to garden their dependencies for updates like this in general? (like, git or not)

I've considered this in the past, but for `reqwest` in particular I was concerned with consumers who had their own uses outside of the generated client i.e. often uses seems to be broader than a single client. What do you think?

progenitor-client/Cargo.toml
lifning: i guess the way to avoid it would be re-implementing them ourselves, which feels pretty far out of scope.
(for what it's worth, we're pulling these in to avoid pulling in all of tungstenite)
this (and potential future such cases of adding new deps) would also be less of an issue if we were to move to emitting code that uses crate re-exports from progenitor-client

lifning:
> > thinking about it more, would it perhaps make sense for us to just re-export everything the generated code needs from our crate, and have the output import them that way so consumers don't have to garden their dependencies for updates like this in general? (like, git or not)
> 
> I've considered this in the past, but for `reqwest` in particular I was concerned with consumers who had their own uses outside of the generated client i.e. often uses seems to be broader than a single client. What do you think?

if they're forced to version-match with us either way, i think it's better to model it within the cargo metadata rather than leaking the dependency graph.  as it sits it feels more awkward than a re-export, especially when it comes to adding new features like this which are now a breaking change with much larger footprint than they'd have otherwise (making them find out which of *our* dependencies to version-bump in *their* Cargo.toml when our macro output stops working due to an update) vs. just having the transitive dependency solved in the golden path where people *don't* have their own independent dependency declared on reqwest (in addition to the comment above about base64 and rand as well).

essentially it would avoid having to explain (and keep re-explaining any time it changes) what to add to Cargo.toml when the general expectation people have when they think "i want to depend on progenitor-client" is to type `progenitor-client = "X.Y.Z"`.

and re-exporting crates is a pretty common pattern in my personal experience - rocket re-exports tokio, for example, and in projects where i've needed to use parts of tokio directly in conjunction, i'm generally happier to put something like `use rocket::tokio;` in my code rather than something like `tokio = "X.Y.Z" # NOTE: must match the version rocket depends on!` in my Cargo.toml (which i've had to do with other crates)

progenitor-impl/src/method.rs
lifning: correct re: it not being application/json - the schema emitted from the API generated by dropshot looks like
```json
"/instance/serial": {
    "get": {
        "operationId": "instance_serial",
        "responses": { "default": { "description": "", "content": { "*/*": { "schema": {} } } } },
        "x-dropshot-websocket": {}
    }
}
```
so `*/*` is what we end up calling it.

ahl:
@lifning ok, let's try it.

ahl:
And by "it" I mean re-exporting reqwest.

progenitor-impl/src/method.rs
lifning: ah! my mistake, we wouldn't, as per below we don't yet describe any error responses in the generated OpenAPI schema.

sample_openapi/propolis-server.json
lifning: presently Dropshot just returns 400 Bad Request for all errors that can occur in its method for performing the websocket upgrade (and technically could also return a 500 Internal Error if a gamma ray flips the right bit, but that code is ordinarily unreachable), and doesn't explicitly specify any of that in the schema as is evident here.

I think the appropriate thing to do would be to update Dropshot to use more specific/appropriate error codes (e.g. 426 Upgrade Required for version mismatch) and model them accordingly in the generated schema for all websocket channel endpoints, at which point progenitor-client's ResponseValue would have something a little more clear to show for what went wrong.

lifning:
updated without re-exports thanks to the newly-tagged reqwest release! https://github.com/seanmonstar/reqwest/releases/tag/v0.11.12

progenitor-impl/src/lib.rs
```diff
@@ -361,6 +366,8 @@ impl Generator {
                     RequestBuilderExt,
                     ResponseValue,
                 };
+                #[allow(unused_imports)]
+                use std::convert::TryInto;
```
luqmana, line 370: Is this needed? All the uses of `.try_into()` we generate are on generic types that explicitly have a `TryInto` trait bound and thus don't require importing the trait.

luqmana (approved):
One small nit re: the unused import but otherwise LGTM. Thanks for driving this forward lif!

I think it'd be worth figuring out a different approach than requiring every consumer to manually add the needed dependencies but I think that's outside the scope of this PR.

progenitor-impl/src/lib.rs
lifning: evidently not! it was in a previous incarnation and i neglected to try removing it again, thanks

PR #200, opened 2022-09-23
update to 1.64 and use workspace dependencies

justinmmott:
Also add rust-toolchain file to ensure nighly version of rustfmt


ahl:
re-open when conflicts are resolved

PR #210, opened 2022-10-06
feat: support header parameters

### LINKED ISSUE #255: Headers not supported
https://github.com/oxidecomputer/progenitor/blob/4e2dcc5/progenitor-impl/src/method.rs#L330
e.g. I have a header which allows telling the backend the API version the client is expecting.
```yaml
  parameters:
    APIVersionParameterHeader:
      name: X-Foo-ApiVersion
      in: header
      required: false
      schema:
        type: string
```

DDtKey:
Closes #255
Base support for Header parameters.

And small additional change: use `Vec::with_capacity` instead of `Vec::new`

All current tests passes, but it would be nice to add some test with `header` params (I tested it on my own examples). 
I've not found `Header` in `dropshot` to write "specific-test". So we can choose any public openapi spec to add. (Upd: see https://github.com/oxidecomputer/progenitor/pull/210#issuecomment-1298794146)

### DIFF
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -111,6 +111,7 @@ enum OperationParameterType {
 enum OperationParameterKind {
     Path,
     Query(bool),
+    Header(bool),
     Body(BodyContentType),
 }
 
@@ -313,6 +314,7 @@ impl Generator {
                             parameter_data.name.clone(),
                             !parameter_data.required,
                         ));
+
                         Ok(OperationParameter {
                             name: sanitize(&parameter_data.name, Case::Snake),
                             api_name: parameter_data.name.clone(),
@@ -323,7 +325,38 @@ impl Generator {
                             ),
                         })
                     }
+                    openapiv3::Parameter::Header {
+                        parameter_data,
+                        style: openapiv3::HeaderStyle::Simple,
+                    } => {
+                        let mut schema = parameter_data.schema()?.to_schema();
+                        let name = sanitize(
+                            &format!(
+                                "{}-{}",
+                                operation.operation_id.as_ref().unwrap(),
+                                &parameter_data.name,
+                            ),
+                            Case::Pascal,
+                        );
 
+                        if !parameter_data.required {
+                            schema = make_optional(schema);
+                        }
+
+                        let typ = self
+                            .type_space
+                            .add_type_with_name(&schema, Some(name))?;
+
+                        Ok(OperationParameter {
+                            name: sanitize(&parameter_data.name, Case::Snake),
+                            api_name: parameter_data.name.clone(),
+                            description: parameter_data.description.clone(),
+                            typ: OperationParameterType::Type(typ),
+                            kind: OperationParameterKind::Header(
+                                parameter_data.required,
+                            ),
+                        })
+                    }
                     openapiv3::Parameter::Path { style, .. } => {
                         Err(Error::UnexpectedFormat(format!(
                             "unsupported style of path parameter {:#?}",
@@ -334,12 +367,6 @@ impl Generator {
                         Err(Error::UnexpectedFormat(format!(
                             "unsupported style of query parameter {:#?}",
                             style,
-                        )))
-                    }
-                    header @ openapiv3::Parameter::Header { .. } => {
-                        Err(Error::UnexpectedFormat(format!(
-                            "header parameters are not supported {:#?}",
-                            header,
                         )))
                     }
                     cookie @ openapiv3::Parameter::Cookie { .. } => {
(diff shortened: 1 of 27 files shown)


DDtKey:
@ahl could you take a look?  (sorry for bothering, i don't know who is maintainer)

ahl:
Thanks for this. I'm going to take a look, but I wanted to set the expectation that it might be a week or two as I'm a little slammed at the moment.

jayvdb:
IMO it would be helpful to add a header to the files in https://github.com/oxidecomputer/progenitor/tree/main/sample_openapi , esp `keeper.json`, and update the https://github.com/oxidecomputer/progenitor/tree/main/example-build if required to make this work.

DDtKey:
@jayvdb I totally agree, but I didn't want to change existed APIs because I'm not sure if they really has any headers as params in their original specs. 

I've checked and found that `keeper` has required `Authorization` header ([link](https://github.com/jclulow/keeper/blob/b32a1f39ed1c54123662cf405e8e0ab3299fc8d2/server/src/main.rs#L69)), so I've added it to samples and updated tests.

jayvdb:
fyi, for non-mandatory headers, I see

```rust
rror[E0277]: `?` couldn't convert the error to `progenitor_client::Error`
   --> clients/foo/src/lib.rs:501:78
    |
501 |             header_map.append("X-Foo-ApiVersion", HeaderValue::try_from(v)?);
    |                                                                              ^ the trait `std::convert::From<reqwest::header::InvalidHeaderValue>` is not implemented for `progenitor_client::Error`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the following other types implement trait `std::convert::From<T>`:
              <progenitor_client::Error<E> as std::convert::From<reqwest::Error>>
    = note: required for `std::result::Result<progenitor_client::ResponseValue<types::ApplicationRegistrationInfo>, progenitor_client::Error>` to implement `std::ops::FromResidual<std::result::Result<std::convert::Infallible, reqwest::header::InvalidHeaderValue>>`
```

jayvdb (approved):
WFM. Thanks.

DDtKey:
Gentle ping @ahl

DDtKey:
Gentle ping @ahl

ahl (commented):
looks good; apologies for the long delay

progenitor-impl/src/method.rs
```diff
@@ -734,6 +765,45 @@ impl Generator {
             (query_build, query_use)
         };
 
+        let headers = method
+            .params
+            .iter()
+            .filter_map(|param| match &param.kind {
+                OperationParameterKind::Header(required) => {
+                    let qn = &param.api_name;
+                    let qn_ident = format_ident!("{}", &param.name);
```
ahl, line 774: I think this stood for "query name", but this is fine...

ahl (changes requested):
can you please add a note to the CHANGELOG.md file?

DDtKey:
> can you please add a note to the CHANGELOG.md file?

Well, I see only [this file](https://github.com/oxidecomputer/progenitor/blob/main/CHANGELOG.adoc) and it tells that it will be updated automatically on release. Isn't? 🤔

progenitor-impl/src/method.rs
DDtKey: Oh, yeah. Changed to `hn` to keep the naming style

ahl:
> > can you please add a note to the CHANGELOG.md file?
> 
> Well, I see only [this file](https://github.com/oxidecomputer/progenitor/blob/main/CHANGELOG.adoc) and it tells that it will be updated automatically on release. Isn't? 🤔

The sections are updated; the content is added by hand; check out the history for the file.

DDtKey:
> > > can you please add a note to the CHANGELOG.md file?
> 
> > 
> 
> > Well, I see only [this file](https://github.com/oxidecomputer/progenitor/blob/main/CHANGELOG.adoc) and it tells that it will be updated automatically on release. Isn't? 🤔
> 
> 
> 
> The sections are updated; the content is added by hand; check out the history for the file.

Oh, got it. Check please.
A bit confusing to be honest, probably only to me.

ahl (approved):
nice

ahl:
`EXPECTORATE=overwrite cargo test`

DDtKey:
> `EXPECTORATE=overwrite cargo test`

Yeah, sorry for some mess here. I'm just not near to my main laptop. So I don't have all of tools. 

It should be fine now, just pls use squashing merge strategy if CI is green again 😅

DDtKey:
@ahl updated, all checks are passed locally (was finally able to return home)

Sorry for the mess one more time

DDtKey:
@ahl could you approve CI runs one more time? I believe we can merge this PR after last changes

ahl:
rerunning a failed macos CI run that appeared to be a transient issue

ahl:
woot! Thanks for taking the long slow walk to get this done @DDtKey!

PR #227, opened 2022-11-07
Add yaml support

jayvdb:


jclulow:
I think it would be better to make the format choice explicit using a flag of some kind (e.g., `-f, --format <FORMAT>`, rather than trying to guess by looking in the file, so that you would then invoke:

```
$ progenitor -f yaml -i your/openapi/document ...
```

The default value for `-f` would be `json` if not specified, for compatibility.

ahl:
My understanding is that yaml is a superset of json--why would we require users to specify the file type if we don't have to?

jclulow:
I think it is better to be strict.  If what I have is a JSON file, I don't want it to be accidentally mis-parsed as YAML if it is not valid for some reason.

jayvdb:
Note other readers often use some sort of trick to automate this, either file extension, looking for `{`, and HTTP `content-type`.
e.g. https://github.com/paperclip-rs/paperclip/blob/43cfea1/src/v2/mod.rs#L110
https://github.com/kstasik/schema-tools/blob/aadd4b5/src/schema.rs#L70
https://github.com/maxdeviant/sdkgen/blob/58db7b41/src/sdkgen/src/main.rs#L56

Having an automated approach becomes more important when implementing https://github.com/oxidecomputer/progenitor/issues/228 , as the external files may be using a different format than the initial file/url specified on the CLI.  e.g. a YAML file may refer to a JSON file.

I suggest the following order for determining the format:
1. an explicit CLI arg,
2. content-type (would exist at this level when https://github.com/oxidecomputer/progenitor/issues/228 is implemented)
3.  file extension,
4. detect using `{`

i.e. the CLI arg would not default to json, but if supplied it overrides all.

ahl:
> I think it is better to be strict.

I think it it sometimes better to be string and sometimes better to be relaxed. Admittedly this is a harder position to navigate.

> If what I believe that I have is a JSON file, I don't want it to be accidentally mis-parsed as YAML if it is not valid for some reason.

What negative consequence would you imagine happening?

progenitor/src/main.rs
```diff
@@ -171,7 +172,18 @@ pub fn load_api<P>(p: P) -> Result<OpenAPI>
 where
     P: AsRef<Path>,
 {
-    let f = File::open(p)?;
-    let api = serde_json::from_reader(f)?;
+    let mut f = File::open(p)?;
+
+    let mut buf = [b' '];
+    while buf[0].is_ascii_whitespace() {
```
ahl, line 178: this doesn't seem right: a json file could start with whitespace, couldn't it?

ahl (changes requested):
If we're doing this for the command, we should do this for the macro.

Off the cuff, I think my preferences for inferring the type of the file would be
1. the file name
2. try it as json, fall back to yaml

I don't love option 1. It's kind of gross, and I have plans for the macro that means that the file name may not be as readily available.

I do like option 2 even though it's inefficient:
- If the file is JSON but there's an error, the serde_yaml error will be basically the same
- serde_json is a lot faster so I don't want to just use serde_yaml for everything even though it would work
- Slowing down the error case for a badly formatted file would be fine.

progenitor/src/main.rs
jayvdb: This loops ignores all leading whitespace in the file, so that the `if` below is checking whether the first non-whitespace character is `{`.

ahl: right! fair enough. Sounds like you prefer this to the options I mentioned?

jayvdb: Well, IMO we need some auto-detection to handle https://github.com/oxidecomputer/progenitor/issues/228 in order to still do the right thing if the content-type isnt recognised , and often raw json & yaml files have strange content-types provided by the server.

ahl: yes, agreed. I suggested that we try to parse as json and then, failing that, parse as yaml. Would that not suffice?

jayvdb: Yea, I like that, and will do that next time I revisit this PR.

progenitor-impl/tests/test_output.rs
```diff
@@ -1,23 +1,32 @@
...
 
 #[track_caller]
 fn verify_apis(openapi_file: &str) {
     let mut in_path = PathBuf::from("../sample_openapi");
-    in_path.push(format!("{}.json", openapi_file));
+    if openapi_file.contains('.') {
+        in_path.push(openapi_file);
+    } else {
+        in_path.push(format!("{}.json", openapi_file));
+    }
```
ahl, line 17: just add .json to the callers

progenitor-macro/src/lib.rs
```diff
@@ -33,8 +34,8 @@ use syn::LitStr;
 /// );
 /// ```
 ///
-/// The `spec` key is required; it is the OpenAPI document from which the
-/// client is derived.
+/// The `spec` key is required; it is the OpenAPI document path from which the
+/// client is derived, and may be either JSON or YAML.
```
ahl, line 38:
```suggestion
/// The `spec` key is required; it is the path of the OpenAPI document (JSON or YAML)  from which the client is derived.
```

progenitor/src/lib.rs
```diff
@@ -11,6 +11,7 @@
 //! README](https://github.com/oxidecomputer/progenitor/README.md)
 
 pub use progenitor_client;
+pub use progenitor_impl::util::load_api;
```
ahl, line 14: I don't think this should be in the progenitor public API

progenitor-impl/src/lib.rs
```diff
@@ -26,6 +26,10 @@ pub enum Error {
     TypeError(#[from] typify::Error),
     #[error("unexpected or unhandled format in the OpenAPI document {0}")]
     UnexpectedFormat(String),
+    #[error("invalid file location {0}")]
+    InvalidFileLocation(String),
+    #[error("unrecognised file contents {0}")]
+    InvalidFileContents(String),
```
ahl, line 32: if load_api isn't public we don't need these error cases. I'd prefer copying the code in both progenitor/src/main.rs and progenitor-macro

At some point, progenitor-macro will switch to something that uses include_str!

ahl (changes requested):
I'd rather have some non-keeper yaml file since otherwise it will be annoying to sync up the json and yaml files as folks add test cases to them

progenitor-impl/src/lib.rs
jayvdb: yup, ok :+1:

ahl (approved):

PR #233, opened 2022-11-09
Add generated crate metadata

jayvdb:
This adds a CLI arg for the crate registry to publish to, and a hardcoded `license` value which at least makes people think about licensing.
Ideally some other approach is found to make the crate metadata more easily customised without resorting to lots of CLI args.

### DIFF
--- a/progenitor/src/main.rs
+++ b/progenitor/src/main.rs
@@ -36,6 +36,12 @@ struct Args {
     /// Target Rust crate version
     #[clap(short = 'v', long)]
     version: String,
+    /// Target Rust crate registry
+    #[clap(long)]
+    registry_name: Option<String>,
+    /// Target crate license
+    #[clap(long, default_value = "SPECIFY A LICENSE BEFORE PUBLISHING")]
+    license_name: String,
 
     /// SDK interface style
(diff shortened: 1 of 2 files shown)


progenitor/src/main.rs
```diff
@@ -25,6 +25,9 @@ struct Args {
     /// Target Rust crate version
     #[clap(short = 'v', long)]
     version: String,
+    /// Target Rust crate repository
+    #[clap(long)]
+    registry_name: String,
```
ahl, line 30: Let's make this optional.

progenitor/src/main.rs
```diff
@@ -123,12 +127,15 @@ fn main() -> Result<()> {
                 name = \"{}\"\n\
                 version = \"{}\"\n\
                 edition = \"2021\"\n\
+                publish = [\"{}\"]\n\
+                license = \"UNLICENSED\"\n\
```
ahl, line 131:
```suggestion
                # license = \"SPECIFY A LICENSE BEFORE PUBLISHING\"\n\
```

ahl (changes requested):

jayvdb:
@ahl , updated

ahl (approved):
looks good

PR #234, opened 2022-11-10
allow for type adjustments: renames, and additional derives

ahl:
This won't build on github until I merge https://github.com/oxidecomputer/typify/pull/131

Thinking about the right interface.

### DIFF
--- a/progenitor/tests/build_nexus.rs
+++ b/progenitor/tests/build_nexus.rs
@@ -57,7 +57,6 @@ mod builder_tagged {
     use futures::StreamExt;
 
     mod nexus_client {
-
         progenitor::generate_api!(
             spec = "../sample_openapi/nexus.json",
             interface = Builder,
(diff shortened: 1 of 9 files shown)


progenitor/tests/build_nexus.rs
```diff
@@ -31,6 +31,11 @@ mod builder_untagged {
             spec = "../sample_openapi/nexus.json",
             interface = Builder,
             tags = Merged,
+            adjustments = {
+                "Name" = {
+                    derives = [Hash],
+                }
+            }
```
ahl, line 38: @andrewjstone this is the syntax I'm thinking about, but would welcome ... something better. I don't want to just do

```rust
per_type_derives = {
    "Name" = [Hash]
}
```

... because I think it makes more sense to lump together various tweaks one might make to a type... although I can't readily think of more...

ahl:
Note, I might change `adjustments` to `patch` or `override`
Also, I should change `"Name"` to just `Name` I think to be consistent with the handling of types/traits

andrewjstone:
> Note, I might change `adjustments` to `patch` or `override` Also, I should change `"Name"` to just `Name` I think to be consistent with the handling of types/traits

Putting it here for posterity: As discussed earlier I like `patch`.

Issue #244, opened 2022-11-18
Generated code contains copyright statement in progenitor_client.rs


jayvdb:
generated `progenitor_client.rs` contains
```
// Copyright 2022 Oxide Computer Company
...
```
This makes it unusable for my work purposes.  The company is happy to have MPL dependencies, but code in our repos needs to be free of other peoples copyrights.

It looks like the generated file is copy of https://github.com/oxidecomputer/progenitor/blob/main/progenitor-client/src/progenitor_client.rs .

Why not make the generated crate depend on https://crates.io/crates/progenitor-client ?

jayvdb:
I guess the reason is that including the source code allows uncommitted/unreleased changes to `progenitor-client` to be accessed by the tests without needing to do tricky fiddling with the generated Cargo.toml to add `progenitor-client` using `path`.

jayvdb:
https://github.com/oxidecomputer/progenitor/pull/246 is one approach.  This is my preference, as it also DRY's up the code a bit if there are multiple generated clients.

A simpler approach is to remove the copyright statement at the top of the "generated" file, but that doesnt completely really solve it because the "generated" file was copied from MPL project, and so it is implicitly MPL unless there is some explicit relicensing done during generation.

ahl:
> Why not make the generated crate depend on https://crates.io/crates/progenitor-client ?

This was an intentional design choice. The project is evolving rapidly with infrequent updates to crates.io. Embedding the code within the generated crate means that we don't need to worry about dependencies in generated code on some support code not yet on crates.io.

I'd suggest you use the builder or macro form rather than the generated crate. If you need to use the generated crate then consider scripting the modifications you need to depend on the crates.io version of progenitor-client.

jayvdb:
> If you need to use the generated crate then consider scripting the modifications you need to depend on the crates.io version of progenitor-client.

https://github.com/oxidecomputer/progenitor/pull/246 adds that possibility as a CLI flag.

PR #246, opened 2022-11-21
allow disabling transclusion

jayvdb:
Resolves https://github.com/oxidecomputer/progenitor/issues/244

This isnt very neat code - more than happy to adjust it to make it less ugly.

### DIFF
--- a/progenitor/Cargo.toml
+++ b/progenitor/Cargo.toml
@@ -22,8 +22,8 @@ serde_yaml = "0.9"
 clap = { version = "4.1.1", features = ["derive"] }
 
 [build-dependencies]
-built = { version = "0.5" }
-project-root = "*"
+built = { version = "0.5", features = ["git2"] }
+project-root = "0.2"
 
 [dev-dependencies]
 base64 = "0.21"
--- a/progenitor/src/main.rs
+++ b/progenitor/src/main.rs
@@ -17,6 +17,11 @@ pub mod built_info {
     include!(concat!(env!("OUT_DIR"), "/built.rs"));
 }
 
+/// Determine if current version is a pre-release or was built from a git-repo
+fn release_is_unstable() -> bool {
+    !built_info::PKG_VERSION_PRE.is_empty() || built_info::GIT_VERSION.is_some()
+}
+
 #[derive(Parser)]
 struct Args {
     /// OpenAPI definition document (JSON or YAML)
@@ -38,6 +43,9 @@ struct Args {
     /// SDK tag style
     #[clap(value_enum, long, default_value_t = TagArg::Merged)]
     tags: TagArg,
+    /// Include client
+    #[clap(default_value = match release_is_unstable() { true => "true", false => "false" }, long, action = clap::ArgAction::Set)]
+    include_client: bool,
 }
 
 #[derive(Copy, Clone, ValueEnum)]
@@ -135,7 +143,7 @@ fn main() -> Result<()> {
                 \n",
                 name,
                 version,
-                dependencies(builder).join("\n"),
+                dependencies(builder, args.include_client).join("\n"),
             );
 
             save(&toml, tomlout.as_str())?;
@@ -150,7 +158,11 @@ fn main() -> Result<()> {
             /*
              * Create the Rust source file containing the generated client:
              */
-            let lib_code = format!("mod progenitor_client;\n\n{}", api_code);
+            let lib_code = if args.include_client {
+                format!("mod progenitor_client;\n\n{}", api_code)
+            } else {
+                api_code
+            };
             let mut librs = src.clone();
             librs.push("lib.rs");
             save(librs, lib_code.as_str())?;
@@ -158,10 +170,12 @@ fn main() -> Result<()> {
             /*
              * Create the Rust source file containing the support code:
              */
-            let progenitor_client_code = progenitor_client::code();
-            let mut clientrs = src;
-            clientrs.push("progenitor_client.rs");
-            save(clientrs, progenitor_client_code)?;
+            if args.include_client {
+                let progenitor_client_code = progenitor_client::code();
+                let mut clientrs = src;
+                clientrs.push("progenitor_client.rs");
+                save(clientrs, progenitor_client_code)?;
+            }
         }
 
         Err(e) => {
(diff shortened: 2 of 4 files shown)


ahl:
How about this:

If `$CARGO_PKG_VERSION_PRE` is set, default to copying the code in; otherwise default to using the crate and do `"progenitor-client = \"{}\"", env!("CARGO_PKG_VERSION")`. Thoughts?

jayvdb:
Sounds good.  Could we still have a CLI arg to allow overriding?

e.g. I am using a hacked version of `progenitor` & `typify` atm, and dont want to be creating a fork on crates.io because I expect my hacks to make it into the repos fairly easily, but I need to be using crates.io `progenitor-client` due to the licensing issue in https://github.com/oxidecomputer/progenitor/issues/244 .  So I would prefer to be using a CLI flag, but I can survive without it.

An alternative approach to the `--transclude` CLI bool flag is inverting it to be a `--use-published-client[=any-valid-dependency-toml]` such as `--use-published-client` as shorthand for using `CARGO_PKG_VERSION`, but then people can also use `--use-published-client='{ git = "https://github.com/their/fork" }'` .  And I could add `--use-published-client=false` or `--no-use-published-client` to allow explicit transclusion.

ahl:
> Sounds good. Could we still have a CLI arg to allow overriding?

That sounds fine; the default can change depending on whether it's a pre-release or not.

> e.g. I am using a hacked version of `progenitor` & `typify` atm, and dont want to be creating a fork on crates.io because I expect my hacks to make it into the repos fairly easily, but I need to be using crates.io `progenitor-client` due to the licensing issue in #244 . So I would prefer to be using a CLI flag, but I can survive without it.

It doesn't seem like there's a licensing issue.

> An alternative approach to the `--transclude` CLI bool flag is inverting it to be a `--use-published-client[=any-valid-dependency-toml]` such as `--use-published-client` as shorthand for using `CARGO_PKG_VERSION`, but then people can also use `--use-published-client='{ git = "https://github.com/their/fork" }'` . And I could add `--use-published-client=false` or `--no-use-published-client` to allow explicit transclusion.

Let's do `--include-client=true/false`

- pre-release: default would be true; setting to false would have a `*` version dependency (and might break on a breaking change to progenitor-client)
- release: default of false with the dependency on the current version.

Please also include documentation updates in this PR.

progenitor/Cargo.toml
```diff
@@ -5,17 +5,22 @@ edition = "2021"
...
 progenitor-macro = { version = "0.2.1-dev", path = "../progenitor-macro" }
 anyhow = "1.0"
 openapiv3 = "1.0.0"
+quote = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 clap = { version = "4.0.29", features = ["derive"] }
 
+[build-dependencies]
+built = { version = "0.5", features = ["git2"] }
```
jayvdb, line 22: We can remove the feature `git2` if we remove the use of `GIT_VERSION`, which is lifted directly from the docs of https://crates.io/crates/built

jayvdb:
> Please also include documentation updates in this PR.

Are you only looking for some notes in `docs/progenitor-client.md` about this addition?

I guess that this mode might also be mentioned in `Static Crate` section in README.md, but IMO it is too complicated to be useful there.

jayvdb:
@ahl , could you provide some feedback on this approach.

progenitor/Cargo.toml
ahl: why is `built` necessary? Is direct use of `evn!("CARGO_PKG_VERSION_PRE")` invalid?

progenitor-impl/src/lib.rs
```diff
@@ -475,6 +491,15 @@ impl Generator {
             "serde = { version = \"1.0\", features = [\"derive\"] }",
             "serde_urlencoded = \"0.7\"",
         ];
+        let client_version_dep: String;
+        match &self.settings.client_version {
+            Some(version) => {
```
ahl, line 496: if let Some(version)

progenitor/src/main.rs
```diff
@@ -32,6 +43,9 @@ struct Args {
     /// SDK tag style
     #[clap(value_enum, long, default_value_t = TagArg::Merged)]
     tags: TagArg,
+    /// Include client
+    #[clap(default_value = match release_is_unstable() { true => "true", false => "false" }, long, action = clap::ArgAction::Set)]
+    include_client: Option<bool>,
 }
```
ahl, line 49: can this not be made simpler?

progenitor/src/main.rs
```diff
@@ -82,9 +96,24 @@ where
...
     let api = load_api(&args.input)?;
+    let include_client = match args.include_client {
+        Some(true) => true,
+        Some(false) => false,
+        None => true,
+    };
+
+    let mut settings = GenerationSettings::default();
+
+    if !include_client {
```
ahl, line 107: comment

progenitor/src/main.rs
```diff
@@ -152,10 +181,17 @@ fn main() -> Result<()> {
             /*
              * Create the Rust source file containing the support code:
```
ahl, line 182: let's just conditionally create this file

progenitor-impl/src/lib.rs
```diff
@@ -255,13 +261,23 @@ impl Generator {
...
+                pub use progenitor_client::{ByteStream, Error, ResponseValue};
+                #[allow(unused_imports)]
+                use progenitor_client::{encode_path, RequestBuilderExt};
+            },
+            Some(_) => quote! {
+                // Re-export ResponseValue and Error since those are used by the
+                // public interface of Client.
+                pub use crate::progenitor_client::{ByteStream, Error, ResponseValue};
+                #[allow(unused_imports)]
+                use crate::progenitor_client::{encode_path, RequestBuilderExt};
```
ahl, line 277: why?

ahl (changes requested):
update https://github.com/oxidecomputer/progenitor#static-crate

ahl:
Let me know if I can help land this.

jayvdb:
I'm ok with updating this.  As https://github.com/oxidecomputer/progenitor/pull/298 is also proposing using `built`, I wanted to get that also into review before returning to this one.

ahl (approved):
looks good

PR #252, opened 2022-11-22
WIP: use reqwest-middleware

jayvdb:
https://github.com/TrueLayer/reqwest-middleware brings retry & tracing support.

I havent updated the docs yet; waiting to see if there is interest in this feature.

I see a bunch of libs in https://github.com/oxidecomputer/third-party-api-clients are using `reqwest-middleware`, and https://github.com/oxidecomputer/reqwest-conditional-middleware

I see https://crates.io/crates/openapitor/0.0.5/dependencies is also using `reqwest-middleware`, but doesnt have its code published on git.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,6 +356,7 @@ dependencies = [
  "progenitor-client",
  "rand",
  "reqwest",
+ "reqwest-middleware",
  "serde",
  "serde_json",
  "uuid",
@@ -368,6 +369,7 @@ dependencies = [
  "chrono",
  "progenitor",
  "reqwest",
+ "reqwest-middleware",
  "schemars",
  "serde",
  "uuid",
@@ -795,6 +797,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2a60c7ce501c71e03a9c9c0d35b861413ae925bd979cc7a4e30d060069aaac8d"
 
 [[package]]
+name = "mime_guess"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4192263c238a5f0d0c6bfd21f336a313a4ce1c450542449ca191bb657b4642ef"
+dependencies = [
+ "mime",
+ "unicase",
+]
+
+[[package]]
 name = "mio"
 version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1042,6 +1054,7 @@ dependencies = [
  "rand",
  "regress",
  "reqwest",
+ "reqwest-middleware",
  "schemars",
  "serde",
  "serde_json",
@@ -1056,6 +1069,7 @@ dependencies = [
  "futures-core",
  "percent-encoding",
  "reqwest",
+ "reqwest-middleware",
  "serde",
  "serde_json",
  "serde_urlencoded",
@@ -1213,6 +1227,7 @@ dependencies = [
  "js-sys",
  "log",
  "mime",
+ "mime_guess",
  "native-tls",
  "once_cell",
  "percent-encoding",
@@ -1229,6 +1244,21 @@ dependencies = [
  "wasm-bindgen-futures",
  "web-sys",
  "winreg",
+]
+
+[[package]]
+name = "reqwest-middleware"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4a1c03e9011a8c59716ad13115550469e081e2e9892656b0ba6a47c907921894"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "http",
+ "reqwest",
+ "serde",
+ "task-local-extensions",
+ "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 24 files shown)


jayvdb:
ping @augustuswm

DDtKey:
I believe it could be optional feature or somehow generalized 🤔
I'd like to use `reqwest-middleware`, but it make sense to keep ability to use pure client as well

jayvdb:
I dont mind making it a CLI-arg opt-in feature.

ahl:
We already let consumers effect the generation with pre- and post-hooks -- would that be sufficient? I'm not a fan of having this by default.. in fact, I've been thinking about ripping out reqwest and moving to a lighter weight HTTP client of our own construction.

jayvdb:
As this requires improvements to `progenitor-client`, it does not seem like this would be easy to do with pre- and post-hooks.

Making the codebase able to support either `reqwest` or `reqwest-middleware` (with `reqwest` as default) will lead the way to allowing other http clients.

ahl:
Thanks for making this, but after looking into reqwest-middleware more, I don't think we're interested in this at this time.

jayvdb:
For anyone interested in using `reqwest_middleware`, the following will transform the generated client.  Sorry it is quite messy.

```rust
// Position of "mod types" in the AST
const MOD_TYPES_INDEX: usize = 3;

/// Transform the `Client`, leaving the struct name as `Client`, and
/// replace `Client` inside the code (which was reqwest::Client) to be
/// `ClientWithMiddleware`.
pub fn use_reqwest_middleware(input: &mut syn::File) {
    /// Removes any `reqwest::` prefix.
    struct RemoveReqwestPrefixVisitor;

    impl VisitMut for RemoveReqwestPrefixVisitor {
        fn visit_path_mut(&mut self, path: &mut syn::Path) {
            let mut new_segments = syn::punctuated::Punctuated::new();
            for segment in path.segments.iter_mut() {
                if segment.ident != "reqwest" {
                    new_segments.push(segment.clone());
                }
            }
            path.segments = new_segments;
            syn::visit_mut::visit_path_mut(self, path);
        }
    }

    // Remove "reqwest::" prefix
    let mut rewriter = RemoveReqwestPrefixVisitor {};
    rewriter.visit_file_mut(input);

    // Used to replace all or some occurrences of `from` to `to`.
    struct Renamer {
        from: String,
        to: syn::Ident,
        found: u64,
        max_times: u64,
    }

    impl Fold for Renamer {
        fn fold_ident(&mut self, id: syn::Ident) -> syn::Ident {
            if id == self.from {
                self.found += 1;
                if self.max_times == 0 || self.found <= self.max_times {
                    self.to.clone()
                } else {
                    id
                }
            } else {
                id
            }
        }
    }

    let mut renamer = Renamer {
        to: proc_macro2::Ident::new("ClientWithMiddleware", proc_macro2::Span::call_site()),
        from: "Client".to_string(),
        max_times: 0,
        found: 0,
    };

    for i in MOD_TYPES_INDEX..input.items.len() - 1 {
        // this renames `pub struct Client` as well as all other mentions of `Client`
        let new_item = input.items[i].clone();
        let new_item = renamer.fold_item(new_item.clone());
        // this bit restores `pub struct Client`
        let mut renamer_revert_once = Renamer {
            to: proc_macro2::Ident::new("Client", proc_macro2::Span::call_site()),
            from: "ClientWithMiddleware".to_string(),
            max_times: 1,
            found: 0,
        };
        let new_item = renamer_revert_once.fold_item(new_item.clone());
        match i.cmp(&(MOD_TYPES_INDEX + 2)) {
            Ordering::Equal => {
                let impl_item: syn::ItemImpl = cast!(&new_item, syn::Item::Impl).to_owned();
                let new_impl_item = transform_client_impl(impl_item);
                input.items[i] = syn::Item::Impl(new_impl_item);
            }
            Ordering::Greater => {
                if let syn::Item::Impl(new_impl) = new_item {
                    let no_doc_test: syn::Attribute = parse_quote! {
                        #[cfg(not(doctest))]
                    };
                    let mut new_impl = new_impl.clone();
                    new_impl.attrs.insert(0, no_doc_test);
                    input.items[i] = syn::Item::Impl(new_impl);
                } else {
                    input.items[i] = new_item;
                }
            }
            Ordering::Less => {
                input.items[i] = new_item;
            }
        }
    }
}

// The following replaces the body of `Client::new` to use reqwest_middleware
fn transform_client_impl(input: syn::ItemImpl) -> syn::ItemImpl {
    let mut new_impl = input;
    let mut new_items: Vec<syn::ImplItem> = vec![];
    for item in new_impl.items.iter_mut() {
        match item {
            syn::ImplItem::Fn(m) => {
                let name = m.sig.ident.to_string();
                if name == "new" {
                    let new_block: syn::Block = parse_quote! {
                        {
                            let dur = std::time::Duration::from_secs(15);

                            let client = reqwest_middleware::ClientBuilder::new(
                                reqwest::ClientBuilder::new()
                                    .connect_timeout(dur)
                                    .timeout(dur)
                                    .build()
                                    .unwrap(),
                            )
                            .build();

                            Self::new_with_client(baseurl, client)
                        }
                    };
                    m.block = new_block;
                }
                new_items.push(syn::ImplItem::Fn(m.to_owned()))
            }
            other => new_items.push(other.to_owned()),
        }
    }
    new_impl.items = new_items;
    new_impl
}
```

After that, re-inject
```rust
use reqwest::{header, Body};
```
and inject
```rust
use reqwest_middleware::ClientWithMiddleware;
```
and in Cargo.toml add
```toml
progenitor-client = { git = "https://github.com/franklin-ai/progenitor/", branch = "ros-enhancements-combined" }
reqwest-middleware = "*"
```
Specifically this is the commit needed for progenitor-client
https://github.com/franklin-ai/progenitor/commit/9573e42ab51f99694189117abdb0ea090efe1cbd

I can tidy this up and publish it as a separate crate if anyone needs it.

jayvdb:
reqwest-middleware enhanced progenitor-client rebased at https://github.com/jayvdb/progenitor/tree/reqwest-middleware-v2

PR #261, opened 2022-11-28
Query param named "query" collision

augustuswm:
If an API spec has a query parameter name `query`, then the generated operation will contain a block that looks like:

**Positional**
```rust
pub async fn operation_method<'a>(
    &'a self,
    query: Option<&'a str>,
) -> Result<ResponseValue<types::ResponseType>, Error<()>> {
    let url = format!(
        "{}/path/to/method",
        self.baseurl,
    );
    let mut query = Vec::new();
    if let Some(v) = &query {
        query.push(("query", v.to_string()));
    }
    // ...
    let request = self.client.get(url).query(&query).build()?;
    // ..
}
```

**Builder**
```rust
pub async fn send(self) -> Result<ResponseValue<types::ResponseType>, Error<()>> {
    let Self {
        client,
        query,
    } = self;
    let query = query.map_err(Error::InvalidRequest)?;
    let url = format!(
        "{}/path/to/method",
        client.baseurl,
    );
    let mut query = Vec::new();
    if let Some(v) = &query {
        query.push(("query", v.to_string()));
    }
    // ...
    let request = client.client.get(url).query(&query).build()?;
    // ...
}
```

This PR initially is here to highlight the collision. I have a different idea for how to resolve it that I will push shortly.

### DIFF
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -728,19 +728,14 @@ impl Generator {
                 _ => None,
             })
             .collect::<Vec<_>>();
-        let (query_build, query_use) = if query_items.is_empty() {
-            (quote! {}, quote! {})
+        let query_build = if query_items.is_empty() {
+            quote! { vec![] }
         } else {
-            let query_build = quote! {
-                let query = [
+            quote! {
+                [
                     #(#query_items,)*
-                ].into_iter().filter_map(|(name, arg)| arg.map(|value| (name, value))).collect::<Vec<_>>();
-            };
-            let query_use = quote! {
-                .query(&query)
-            };
-
-            (query_build, query_use)
+                ]
+            }
         };
 
         let websock_hdrs = if method.dropshot_websocket {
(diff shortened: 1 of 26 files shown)


augustuswm:
This should allow for query or path parameters to be named `query` or `url`. If a parameter is named `client` though (and the builder style is used) we will still run into conflicts.

ahl (commented):
The generic problem is that we're not using hygienic names... maybe we should add a `__progenitor_` prefix to internal variable names.

augustuswm:
Exactly, normally that is the path I would take. The downside would be that it means `__progenitor_` (or any prefix) will appear in the final generated client source code (which I was seeing if we could avoid initially). If this is an ok trade-off to make, then prefixing is likely the way to go.

`client` is a more problematic case though, and I'm stuck on thinking of a solution other than prefixing.

inickles:
https://github.com/oxidecomputer/progenitor/pull/599 was a similar (and updated) PR to solve the same underlying issue and has now been merged.

PR #262, opened 2022-11-29
Path level parameters

augustuswm:
* Adds support for path-level parameters

Operation level parameters override path level parameters when they share the same name based on my understanding of https://spec.openapis.org/oas/latest.html#fixed-fields-7 .

This implementation creates a new map of path level parameters per operation which is wasteful, but outside of this let me know if I'm misunderstanding how parameters should be handled.


ahl (commented):
looks good. thanks

ahl:
I think this will obviate #75

PR #278, opened 2022-12-21
Add support for wasm

dmeijboom:
This PR adds support for wasm when using the `build.rs` flow (using the macro is untested and might not `work). Since `progenitor-client` uses the `stream` feature of `reqwest` it also depends on: https://github.com/seanmonstar/reqwest/issues/1704

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -60,3 +60,29 @@ jobs:
       run: cargo build --locked --tests --verbose
     - name: Run tests
       run: cargo test --locked --verbose
+
+  test-wasm:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install nightly
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: nightly
+        components: rustfmt
+        default: false
+        target: wasm32-unknown-unknown
+    - uses: nanasess/setup-chromedriver@v1
+    - name: Install stable
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        components: rustfmt
+        default: true
+        target: wasm32-unknown-unknown
+    - name: Install wasm-bindgen-cli
+      run: cargo install --version 0.2.84 wasm-bindgen-cli
+    - name: Run tests
+      run: |
+        cd example-wasm
+        CHROMEDRIVER=$(which chromedriver) cargo test --target wasm32-unknown-unknown --locked --verbose
--- a/example-wasm/Cargo.toml
+++ b/example-wasm/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "example-wasm"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+chrono = { version = "0.4", features = ["serde"] }
+progenitor-client = { path = "../progenitor-client" }
+reqwest = { version = "0.11.16", features = ["json", "stream"] }
+base64 = "0.21"
+serde = { version = "1.0", features = ["derive"] }
+uuid = { version = "1.3", features = ["serde", "v4", "js"] }
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3.34"
+
+[build-dependencies]
+progenitor = { path = "../progenitor" }
+serde_json = "1.0"
--- a/progenitor-client/src/progenitor_client.rs
+++ b/progenitor-client/src/progenitor_client.rs
@@ -11,8 +11,13 @@ use futures_core::Stream;
 use reqwest::RequestBuilder;
 use serde::{de::DeserializeOwned, Serialize};
 
+#[cfg(not(target_arch = "wasm32"))]
 type InnerByteStream =
     std::pin::Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>> + Send + Sync>>;
+
+#[cfg(target_arch = "wasm32")]
+type InnerByteStream =
+    std::pin::Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>>>>;
 
 /// Untyped byte stream used for both success and error responses.
 pub struct ByteStream(InnerByteStream);
@@ -76,6 +81,7 @@ impl<T: DeserializeOwned> ResponseValue<T> {
     }
 }
 
+#[cfg(not(target_arch = "wasm32"))]
 impl ResponseValue<reqwest::Upgraded> {
     #[doc(hidden)]
     pub async fn upgrade<E: std::fmt::Debug>(
--- a/progenitor-impl/src/lib.rs
+++ b/progenitor-impl/src/lib.rs
@@ -345,13 +345,18 @@ impl Generator {
                     baseurl: &str,
                     #inner_parameter
                 ) -> Self {
-                    let dur = std::time::Duration::from_secs(15);
-                    let client = reqwest::ClientBuilder::new()
-                        .connect_timeout(dur)
-                        .timeout(dur)
-                        .build()
-                        .unwrap();
-                    Self::new_with_client(baseurl, client, #inner_value)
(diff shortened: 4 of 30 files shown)


ahl:
This is neat. Thank you for doing this. Is it possible to do any testing of the wasm target in CI?

dmeijboom:
> This is neat. Thank you for doing this. Is it possible to do any testing of the wasm target in CI?

Yes, definitely! However, it's blocked by support for the `stream` feature in Reqwest when using wasm. When that's merged I'll look at the tests.

progenitor-impl/src/lib.rs
```diff
@@ -259,13 +259,18 @@ impl Generator {
...
                 ) -> Self {
-                    let dur = std::time::Duration::from_secs(15);
-                    let client = reqwest::ClientBuilder::new()
-                        .connect_timeout(dur)
-                        .timeout(dur)
-                        .build()
-                        .unwrap();
-                    Self::new_with_client(baseurl, client, #inner_value)
+                    #[cfg(target_arch = "wasm32")]
+                    let client = reqwest::ClientBuilder::new();
```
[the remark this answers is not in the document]
dmeijboom: Both `connect_timeout(...)` and `timeout(...)` aren't supported in wasm with reqwest.

progenitor-client/src/progenitor_client.rs
```diff
@@ -76,6 +81,7 @@ impl<T: DeserializeOwned> ResponseValue<T> {
     }
 }
 
+#[cfg(not(target_arch = "wasm32"))]
```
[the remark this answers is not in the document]
dmeijboom: The type `reqwest::Upgraded` doesn't exist in reqwest on wasm. So I assume it's not supported (yet?).

progenitor-client/src/progenitor_client.rs
```diff
@@ -11,9 +11,14 @@ use futures_core::Stream;
 use reqwest::RequestBuilder;
 use serde::{de::DeserializeOwned, Serialize};
 
+#[cfg(not(target_arch = "wasm32"))]
 type InnerByteStream =
     std::pin::Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>> + Send + Sync>>;
 
+#[cfg(target_arch = "wasm32")]
+type InnerByteStream =
+    std::pin::Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>>>>;
```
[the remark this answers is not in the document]
dmeijboom: I know that a `Future` translates to a `Promise` (https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Promise.html) in wasm32 which is `!Send`. I can't find the rationale behind this but I assume it has something to do with the browser runtime which is single threaded (unless you are using web workers).

ahl:
@dmeijboom looks like fixture data needs some love

dmeijboom:
> @dmeijboom looks like fixture data needs some love

Yeah, indeed. I want to pick this MR up after `reqwest` supports the stream feature. It seems like it's moving in the right direction so I hope to continue working on this MR soon!

progenitor-client/src/progenitor_client.rs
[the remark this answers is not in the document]
dmeijboom: I removed `Send + Sync` entirely (also for non-wasm targets) and it seems to compile. Can it be safely removed or was there a reason for adding both markers to the `InnerByteStream` type?

ahl:
this might be unblocked now by #316

dmeijboom:
> this might be unblocked now by #316

Cool :) I'll update the MR later this week!

dmeijboom:
Sorry for the delay 😅 I rebased the PR and updated the fixtures.

dmeijboom:
> Is it possible to do any testing of the wasm target in CI?
> Some CI for the wasm target should now be possible.

I added a basic test case for wasm. It does nothing more than creating a client but at least we can guarantee that clients generated by progenitor are going to compile on wasm. Not sure where to go from here so I'm up for suggestions whether to add more test cases.

example-build/tests/wasm.rs
```diff
@@ -0,0 +1,11 @@
...
+
+wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
+include!(concat!(env!("OUT_DIR"), "/codegen.rs"));
+
+#[wasm_bindgen_test]
+fn test_client_new() {
+    let client = Client::new("http://foo/bar");
+    assert!(client.baseurl == "http://foo/bar");
+}
```
ahl, line 11: the purpose of the examples is for people to copy to them; let's keep the tests out of here.

progenitor-client/src/progenitor_client.rs
```diff
@@ -11,6 +11,11 @@ use futures_core::Stream;
 use reqwest::RequestBuilder;
 use serde::{de::DeserializeOwned, Serialize};
 
+#[cfg(target_arch = "wasm32")]
+type InnerByteStream =
+    std::pin::Pin<Box<dyn Stream<Item = reqwest::Result<Bytes>>>>;
+
```
ahl, line 17: let's move this below the non-wasm one

progenitor-impl/src/lib.rs
```diff
@@ -345,13 +345,18 @@ impl Generator {
...
```
ahl, line 349: ditto; move this below non-wasm

ahl (commented):
looking good; let's move that test and fix a couple of nits and it should be good to go

.github/workflows/rust.yml
```diff
@@ -50,13 +50,20 @@ jobs:
         toolchain: nightly
         components: rustfmt
         default: false
+        target: wasm32-unknown-unknown
```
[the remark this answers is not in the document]
dmeijboom: Yeah indeed, only Linux should be fine

example-build/tests/wasm.rs
dmeijboom: I agree, moved to `example-wasm`!

dmeijboom:
> looking good; let's move that test and fix a couple of nits and it should be good to go

Done! I created a new project called `example-wasm` as I don't want to pollute the other crates with wasm-bindgen-test stuff.

example-wasm/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "example-wasm"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
ahl, line 6: delete

example-build/Cargo.toml
```diff
@@ -9,7 +9,6 @@ chrono = { version = "0.4", features = ["serde"] }
 progenitor-client = { path = "../progenitor-client" }
 reqwest = { version = "0.11.16", features = ["json", "stream"] }
 base64 = "0.21"
-rand = "0.8"
```
ahl, line 12: let's leave it in? clients may need this... even if this one doesn't

ahl (changes requested):
nit

dmeijboom:
Also fixed the wasm tests! Forgot to setup chromedriver.

ahl:
CI is still failing

dmeijboom:
> CI is still failing

The `wasm-bindgen-test-runner` binary was missing. Should be fixed. There isn't an easy way to test GitHub actions unfortunately so I hope/think it works but wasn't able to validate it.

jayvdb:
https://github.com/nektos/act is a nice tool for running github actions locally - often the workflows need a bit of fiddling to make it feasible to run locally.  I havent tried it yet with this project.

ahl:
Let's gooo!

I'm excited to have this; thank you for the contribution. I'm optimistic about this being easy to maintain; and hope you'll be around to help us if we hit any snags.

PR #232, opened 2022-11-09
Disable default features of reqwest and chrono

jayvdb:
I needed to do this in order to avoid cargo-deny complaining about the generated crate.

PR #298, opened 2023-01-09
generate using built dependency versions

jayvdb:
This avoids the generated Cargo.toml having different versions than the project manifests, which are being regularly updated using [dependabot](https://github.com/apps/dependabot).

e.g. at the moment, reqwest is different https://github.com/oxidecomputer/progenitor/blob/57031d7/progenitor/Cargo.toml#L29 (0.11.13) vs https://github.com/oxidecomputer/progenitor/blob/ba189e6/progenitor-impl/src/lib.rs#L500 (0.11.12).

That difference isnt too bad, but this PR is to prevent more serious differences.

Using https://docs.rs/built/ was easiest way to do this.  Happy to re-implement the various bits needed to achieve this with fewer added dependencies, but there are not many - `built` only depends on https://docs.rs/cargo-lock , which has only a few deps which might be omittable with a hand-crafted implementation.

If we dont use the `Cargo.lock`, and instead use `Cargo.toml`, the number/size of deps likely increases due to the complexity of cargo toml parsing libraries.

I avoided adding `built` to `progenitor-impl` to avoid that growing in size, which means if we need this in the macro, the implementation in the progenitor bin needs to be copied across to the -macro. Alternatively we can just add this to `progenitor-impl`, which means the version mapping no longer needs to be passed through the `GenerationSettings` and the fallback default of version `"*"` is no longer needed.

### DIFF
--- a/progenitor-impl/src/lib.rs
+++ b/progenitor-impl/src/lib.rs
@@ -505,43 +505,17 @@ impl Generator {
         })
     }
 
-    pub fn dependencies(&self) -> Vec<String> {
-        let mut deps = vec![
-            "bytes = \"1.3.0\"",
-            "futures-core = \"0.3\"",
-            "percent-encoding = \"2.2\"",
-            "reqwest = { version = \"0.11.13\", default-features=false, features = [\"json\", \"stream\"] }",
-            "serde = { version = \"1.0\", features = [\"derive\"] }",
-            "serde_urlencoded = \"0.7.1\"",
-        ];
-        if self.type_space.uses_regress() {
-            deps.push("regress = \"0.4.1\"")
-        }
-        if self.type_space.uses_uuid() {
-            deps.push(
-                "uuid = { version = \">=0.8.0, <2.0.0\", features = [\"serde\", \"v4\"] }",
-            )
-        }
-        if self.type_space.uses_chrono() {
-            deps.push("chrono = { version = \"0.4\", default-features=false, features = [\"serde\"] }")
-        }
-        if self.uses_futures {
-            deps.push("futures = \"0.3\"")
-        }
-        if self.uses_websockets {
-            deps.push("base64 = \"0.21\"");
-            deps.push("rand = \"0.8\"");
-        }
-        if self.type_space.uses_serde_json() {
-            deps.push("serde_json = \"1.0\"")
-        }
-        deps.sort_unstable();
-        deps.iter().map(ToString::to_string).collect()
-    }
-
     // TODO deprecate?
     pub fn get_type_space(&self) -> &TypeSpace {
         &self.type_space
+    }
+
+    pub fn uses_futures(&self) -> bool {
+        self.uses_futures
+    }
+
+    pub fn uses_websockets(&self) -> bool {
+        self.uses_websockets
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,6 +103,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "built"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5b9c056b9ed43aee5e064b683aa1ec783e19c6acec7559e3ae931b7490472fbe"
+dependencies = [
+ "cargo-lock",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


jayvdb:
https://github.com/rust-secure-code/cargo-auditable would be a better way to store the dep list, but this would only be usable from the `progenitor` binary, or any other binary built using `cargo-auditable` (i.e. the macro would support this, but only if the macro "user" also used `cargo-auditable` when building their binaries.

ahl:
Good idea. It looks like there's some commented code in here; please let me know when you'd like me to take a look.

jayvdb:
@ahl, at the moment, I would appreciate feedback on whether using `built` is ok, or switching to `cargo-auditable` would be better, or if you want some hand-coded slightly more optimal implementation.

ahl:
I have no objection to using `built`

jayvdb:
@ahl, now ready.

progenitor-impl/src/lib.rs
```diff
@@ -492,38 +501,58 @@ impl Generator {
         })
     }
 
+    fn get_dep_version(&self, name: &str) -> String {
+        match self.settings.dependency_versions.get(&name.to_string()) {
+            Some(o) => o.to_string(),
+            None => "*".to_string(),
```
ahl, line 507: it seems like encountering a `None` here would be reflective of a bug, no?

progenitor-macro/src/lib.rs
```diff
@@ -174,7 +179,14 @@ fn do_generate_api(item: TokenStream) -> Result<TokenStream, syn::Error> {
             patch,
             convert,
         } = serde_tokenstream::from_tokenstream(&item.into())?;
+
+        let deps: HashMap<String, String> = built_info::DEPENDENCIES
+            .iter()
+            .map(|(name, version)| (name.to_string(), version.to_string()))
+            .collect();
+
```
ahl, line 187: why is this useful for the macro form?

progenitor-impl/tests/test_specific.rs
```diff
@@ -167,3 +167,36 @@ fn test_default_params() {
         &output,
     );
 }
+
+/// Test generated dependencies are correct
+#[test]
```
ahl, line 172: this file has tests of specify types; I'd suggest that these tests don't belong here.

progenitor-impl/tests/test_specific.rs
```diff
@@ -167,3 +167,36 @@ fn test_default_params() {
...
+fn test_explicit_dependency_versions() {
+    let mut deps: HashMap<String, String> = HashMap::new();
+    deps.insert("bytes".to_string(), "1.2.3".to_string());
+    deps.insert("reqwest".to_string(), "4.5.6".to_string());
+    let mut api = ApiDescription::new();
+    api.register(default_params).unwrap();
+
+    let generator = Generator::new(
+        &GenerationSettings::default().with_dependency_versions(deps),
```
ahl, line 181: clippy says no & needed

progenitor-impl/src/lib.rs
```diff
@@ -492,38 +501,58 @@ impl Generator {
...
+        }
+    }
+
     pub fn dependencies(&self) -> Vec<String> {
```
ahl, line 511: what if we change this to just enumerate a list of dependencies without version numbers and then leave it to the consumer (i.e. progenitor stand-alone) to fill in the version numbers? Then we can do away with any knowledge of version numbers in progenitor-impl (or progenitor-macro) where it really isn't relevant.

ahl (changes requested):

progenitor-impl/src/lib.rs
jayvdb: No. It means that `progenitor-impl` is being used without `.with_dependency_versions` being called, or not provided with an entry for one of the crates.

progenitor-macro/src/lib.rs
jayvdb: yea, the dependencies are not used by the macro. happy to remove.
alternatively we could emit a dependency block that is used by tools like https://github.com/reitermarkus/cargo-eval and others.
```
//! ```cargo
//! ...
//! ```
```

progenitor-impl/tests/test_specific.rs
jayvdb: maybe put these in a new file `test_deps.rs` , or broader `test_cargo_toml.rs` ?

progenitor-impl/src/lib.rs
ahl: I don't want to feature creep this PR any more than necessary. Let's just let the features for reqwest etc be implicit and we can sort out this interaction later.

progenitor-impl/tests/test_specific.rs
ahl: Sure

progenitor-impl/src/lib.rs
jayvdb: That would be a regression of https://github.com/oxidecomputer/progenitor/pull/232 . I'd rather we sort it out this interaction now.

ahl: I'm not sure how that follows, but, ok. You're proposing we replace `dependencies -> Vec` with `uses_X -> bool` for the various optional dependencies? Fine.

ahl (approved):
Looks great!

PR #301, opened 2023-01-12
fix clippy issues

jayvdb:


ahl:
I don't encourage clippy-only changes in particular because
1. I don't think clippy is always right
2. Absent clippy in CI, PRs like this for clippy cleanliness will regress
3. Spending the time to configure clippy and annotate the code for exceptions is not worth it.

If there are particular changes you think are beneficial, let me know. Otherwise, let's close this PR.

jayvdb:
I believe the latter two of these are correct and appropriate changes.

The first one, using `.unwrap_or_else()` due to https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call , is also more correct, but is not beneficial to the end user, and arguably makes the code more complicated.

ahl:
Ok, let's remove that first one.

jayvdb:
@ahl, done

PR #302, opened 2023-01-12
Add method for referring to Client's inner type

bnaecker:


progenitor-impl/src/lib.rs
```diff
@@ -239,6 +239,18 @@ impl Generator {
...
+        let maybe_as_inner = match &self.settings.inner_type {
+            Some(inner) => quote! {
+                /// Return a reference to the inner type stored in `self`.
+                pub fn as_inner(&self) -> &#inner {
+                    &self.inner
+                }
+            },
+            None => quote! {},
+        };
+
```
ahl, line 253:
```suggestion
        // Generate an implementation of a `Self::as_inner` method, if an inner
        // type is defined.
        let maybe_as_inner = self.settings.inner_type.as_ref().map(|inner| {
            quote! {
                /// Return a reference to the inner type stored in `self`.
                pub fn as_inner(&self) -> &#inner {
                    &self.inner
                }
            }
        });
```

To me `as_inner` implies... a conversion? Would I be exposing myself as a Java ex-pat to suggest that this might be `get_inner`?

bnaecker: Yeah, I think that was based on a suggestion from Cliff, but one that doesn't really apply here. It could just be `inner()` actually.

ahl (approved):

PR #303, opened 2023-01-13
documentation examples sometimes are not formatted properly

ahl:


progenitor-impl/tests/output/nexus-builder-tagged.out
```diff
@@ -11167,7 +11166,9 @@ pub trait ClientDisksExt {
     ///
     ///Sends a `GET` request to
     /// `/organizations/{organization_name}/projects/{project_name}/disks/
-    /// {disk_name}` ```ignore
```
ahl, line 11170: examples of the error

PR #304, opened 2023-01-13
slashes are not properly %-encoded in path parameters

ahl:
CLU-LOOOWWWW


bnaecker (approved):
LGTM, thanks! So I understand, is there a reason not to use the [big hammer](https://docs.rs/percent-encoding/latest/percent_encoding/constant.NON_ALPHANUMERIC.html)?

ahl:
> LGTM, thanks! So I understand, is there a reason not to use the [big hammer](https://docs.rs/percent-encoding/latest/percent_encoding/constant.NON_ALPHANUMERIC.html)?

I don't know: I just did what the spec told me (i.e. based on my understanding of the spec)

PR #317, opened 2023-01-24
Let consumers replace types by name

ahl:


bnaecker (commented):
Can't comment too much on the code, but I just tried this out in an autogenerated client for Dendrite. As promised, the replacement type comes through in the `cargo expand` output, and the generated API methods to in fact take the replacement type straight-up! Does what it says on the tin!

PR #345, opened 2023-02-23
progenitor_impl using only typify_impl

jayvdb:
Users of `progenitor_impl` shouldnt need all of `typify` - they only need to depend on `typify_impl`.


ahl:
While this is valid and more efficient, I would prefer consumers rely on typify rather than its dependencies. If there's a significant performance gain to be had, my preference would be to make `typify-macro` dependent on a (default) feature.

PR #349, opened 2023-03-03
initial cut of CLI generation

ahl:
This is a prototype feature, not ready to be released, but I'd rather have it in the repo than maintaining it in a branch. It shouldn't impact consumers that don't explicitly use it.

Closes https://github.com/oxidecomputer/oxide-sdk-and-cli/issues/46

cc: @david-crespo, @zephraph, @karencfv


progenitor-impl/src/cli.rs
```diff
@@ -0,0 +1,482 @@
...
+
+            quote! {
+                #(
+                    .arg(#body_arg)
+                )*
+            }
+        });
+
+        // TODO parameter for body as input json (--body-input?)
+        // TODO parameter to output a body template (--body-template?)
```
karencfv, line 279: By body template do you mean a table?

karencfv (commented):
woohoo!!!! 🎉 Thanks for working on this!

progenitor-impl/src/cli.rs
```diff
@@ -0,0 +1,482 @@
...
+            };
+
+            quote! {
+                #(
+                    .arg(#body_arg)
+                )*
+            }
+        });
+
+        // TODO parameter for body as input json (--body-input?)
```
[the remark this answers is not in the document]
david-crespo: I suspect _we_ need it in the first iteration of the CLI at least as much as the customer would.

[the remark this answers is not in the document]
karencfv: it'd be necessary for the `$ oxide api` command, not so much for the others imo

progenitor-impl/src/cli.rs
```diff
@@ -0,0 +1,552 @@
...
+
+impl Generator {
+    pub fn cli_text(
+        &mut self,
+        spec: &OpenAPI,
+        crate_name: &str,
+    ) -> Result<String> {
+        let output = self.cli(spec, crate_name)?;
+
+        let content = rustfmt_wrapper::rustfmt_config(
```
jayvdb, line 34: according to https://github.com/oxidecomputer/progenitor/issues/362 , wouldnt `rustfmt` not be used in the -impl ?

progenitor-impl/src/cli.rs
ahl: I'd like to have a parameter that outputs "blank" JSON output so that the user can output that to a file, edit it to fill in appropriate values, and then input it to another option that accepts a JSON-formatted file as input.

progenitor-impl/src/cli.rs
ahl: Good point; we'll work this out.

Issue #362, opened 2023-03-17
remove implicit dependency for transitive consumers on `rustfmt` installation


ahl:
See https://github.com/oxidecomputer/typify/pull/221

In short:
- `build.rs` files should use prettyplease
- anonymous code should be unformatted
- stand-alone generation such as xtask or the progenitor binary should use rustfmt-wrapper

PR #368, opened 2023-03-18
Remove rustfmt from progenitor-impl

jayvdb:
Closes https://github.com/oxidecomputer/progenitor/issues/362

### DIFF
--- a/progenitor-impl/src/cli.rs
+++ b/progenitor-impl/src/cli.rs
@@ -10,7 +10,6 @@ use crate::{
     method::{
         OperationParameterKind, OperationParameterType, OperationResponseStatus,
     },
-    space_out_items,
     to_schema::ToSchema,
     util::{sanitize, Case},
     validate_openapi, Generator, Result,
@@ -23,25 +22,6 @@ struct CliOperation {
 }
 
 impl Generator {
-    pub fn cli_text(
-        &mut self,
-        spec: &OpenAPI,
-        crate_name: &str,
-    ) -> Result<String> {
-        let output = self.cli(spec, crate_name)?;
-
-        let content = rustfmt_wrapper::rustfmt_config(
-            rustfmt_wrapper::config::Config {
-                format_strings: Some(true),
-                ..Default::default()
-            },
-            output,
-        )
-        .unwrap();
-
-        space_out_items(content)
-    }
-
     /// Generate a `clap`-based CLI.
     pub fn cli(
         &mut self,
--- a/progenitor-impl/src/lib.rs
+++ b/progenitor-impl/src/lib.rs
@@ -506,41 +506,7 @@ impl Generator {
 
     /// Render text output.
     pub fn generate_text(&mut self, spec: &OpenAPI) -> Result<String> {
-        self.generate_text_impl(
-            spec,
-            rustfmt_wrapper::config::Config::default(),
-        )
-    }
-
-    /// Render text output and normalize doc comments
-    ///
-    /// Requires a nightly install of `rustfmt` (even if the target project is
-    /// not using nightly).
-    pub fn generate_text_normalize_comments(
-        &mut self,
-        spec: &OpenAPI,
-    ) -> Result<String> {
-        self.generate_text_impl(
-            spec,
-            rustfmt_wrapper::config::Config {
-                normalize_doc_attributes: Some(true),
-                wrap_comments: Some(true),
-                ..Default::default()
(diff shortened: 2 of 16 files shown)


progenitor-impl/src/lib.rs
```diff
@@ -487,48 +487,19 @@ impl Generator {
...
-        config: rustfmt_wrapper::config::Config,
-    ) -> Result<String> {
-        let output = self.generate_tokens(spec)?;
-
-        // Format the file with rustfmt.
-        let content = rustfmt_wrapper::rustfmt_config(config, output).unwrap();
-
-        // Add newlines after end-braces at <= two levels of indentation.
-        Ok(if cfg!(not(windows)) {
-            let regex = regex::Regex::new(r#"(})(\n\s{0,8}[^} ])"#).unwrap();
```
jayvdb, line 526: without the pre-formatting, this regex doesnt work.  
It was a nice touch.
Looks like there is nothing in the rustfmt pipeline to help with this https://rust-lang.github.io/rustfmt/?version=master&search=blank .

An alternative would be to look for pub struct/enum, impl, etc, but attributes make that a bit tricky.

Or maybe this new-line based regex could be provided as a utility function which is demonstrated being used after formatting in the README/example/etc?

jayvdb:
If this is what you had in mind, I'll update the README etc.

progenitor-impl/src/cli.rs
```diff
@@ -30,16 +30,7 @@ impl Generator {
     ) -> Result<String> {
         let output = self.cli(spec, crate_name)?;
```
ahl, line 31: let's just delete `cli_text`

progenitor-impl/src/lib.rs
```diff
@@ -505,41 +505,19 @@ impl Generator {
 
     /// Render text output.
```
ahl, line 506: I think we should just delete generate_text and generatex_text_normalize_comments

progenitor-impl/tests/test_output.rs
```diff
@@ -26,6 +26,16 @@ where
     }
 }
 
+fn generate_formatted(generator: &mut Generator, spec: &OpenAPI) -> String {
```
ahl, line 29: Let's make the goal to have no changes in the output fixtures

progenitor/src/main.rs
```diff
@@ -84,6 +84,15 @@ impl From<TagArg> for TagStyle {
     }
 }
 
+fn reformat_code(input: String) -> String {
+    let config = rustfmt_wrapper::config::Config {
+        normalize_doc_attributes: Some(true),
+        wrap_comments: Some(true),
+        ..Default::default()
+    };
+    rustfmt_wrapper::rustfmt_config(config, input).unwrap()
```
ahl, line 93: add the item spacing here?

ahl (commented):
good direction; sorry it took me so long to get to this.

progenitor-impl/src/lib.rs
jayvdb: `generate_text` has been in the `README.md` sample `build.rs`.

progenitor-impl/tests/test_output.rs
jayvdb: There is still one changed fixture - `progenitor-impl/tests/output/nexus-cli.out` - this is because the removed formatting code in cli.rs had different rustfmt settings, specifically it used only `format_strings: Some(true),` which wasnt used in lib.rs.

progenitor-impl/src/lib.rs
ahl: still! let's get rid of them!

progenitor-impl/tests/test_output.rs
ahl: What's it look like if we `format_strings: Some(true)` for everything?

ahl (approved):
looks good; one outstanding item for you to consider; one conflict to resolve and I'll merge it

progenitor-impl/tests/test_output.rs
jayvdb: IMO it looks great - it does however change the non-cli .out files, so was optional which approach to take.  I'll add it and you can judge for yourself, but IMO it is worth the extra changes.

jayvdb:
ready to roll @ahl

ahl:
Thanks for doing this!

PR #430, opened 2023-04-21
Allow for parameters in content types

DouglasDwyer:
Right now, any component schemas with content types such as `application/json; charset=utf-8` aren't identified correctly by the client generator. This is problematic, because it leads to endpoints being generated that return `ResponseValue<ByteStream>` rather than the deserialized component type. To fix this, I've added some simple logic that ignores anything beyond a `;` in the content type.

### DIFF
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -456,8 +456,11 @@ impl Generator {
                 // content type of the response just as it currently examines
                 // the status code.
                 let typ = if let Some(mt) =
-                    response.content.iter().find_map(|(x, v)| (x == "application/json" || x.starts_with("application/json;")).then_some(v))
-                {
+                    response.content.iter().find_map(|(x, v)| {
+                        (x == "application/json"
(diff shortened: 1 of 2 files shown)


ahl (approved):
looks good; thanks

Is there an OpenAPI document in particular where you're seeing this failure?

DouglasDwyer:
@ahl yes, I'm generating a client for a server written using [poem-openapi](https://crates.io/crates/poem-openapi). The autogenerated output spec contains `application/json; charset=utf-8` types :)

ahl:
One more thing: would you mind adding a test to validate this? You could change keeper.json or buildomat.json.

DouglasDwyer:
Sure! I've gone ahead and swapped out the content types in buildomat.json to be `application/json; charset=utf-8`. Let me know if there's anything more that I need to do. I believe that the existing tests correctly cover this case - when run against the updated buildomat.json, tests fail on the repository's main branch but not on this PR's branch.

ahl (approved):
looks good; thanks for adding the test.

PR #436, opened 2023-04-24
add accept header

### LINKED ISSUE #375: `Accept` header is not set
My API returns either XML or JSON based on the Accept header. It defaults to XML and since this library doesn't set the `Accept` header for requests all my requests fail because it can't parse the XML.

ahl:
Fixes #375 

Note that we don't add the header for responses that are raw or upgrade.

### DIFF
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -999,16 +999,37 @@ impl Generator {
                         }
                     }
                     OperationResponseType::Upgrade => {
-                        if response.status_code == OperationResponseStatus::Default {
-                            return quote! { } // catch-all handled below
+                        if response.status_code
+                            == OperationResponseStatus::Default
+                        {
+                            return quote! {}; // catch-all handled below
                         } else {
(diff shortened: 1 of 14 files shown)


progenitor-impl/src/method.rs
```diff
@@ -999,17 +999,38 @@ impl Generator {
...
-                        if response.status_code == OperationResponseStatus::Default {
-                            return quote! { } // catch-all handled below
+                        if response.status_code
+                            == OperationResponseStatus::Default
+                        {
+                            return quote! {}; // catch-all handled below
                         } else {
-                            todo!("non-default error response handling for upgrade requests is not yet implemented");
+                            todo!(
+                                "non-default error response handling for \
```
ahl, line 1008: this long line was causing rustfmt to give up

augustuswm (approved):
Makes sense to me, and tests pass after running expectorate.

ahl:
> Makes sense to me, and tests pass after running expectorate.

d'oh; fixed

PR #440, opened 2023-04-25
remove generate_text

ahl:
This is breaking, and may break things, but I'm convinced that we just want to generate code and let the consumers figure out the right way to format it.


truh:
Please update the readme. One of the examples uses `generate_text`

Issue #344, opened 2023-02-21
Multiple response types not supported


jayvdb:
It is not possible currently to have two response types due to the assert at
https://github.com/oxidecomputer/progenitor/blob/1b942d1/progenitor-impl/src/method.rs#L1118-L1120

I wonder if it is easier and acceptable to slightly improve the situation by allowing one response body type, and one or more additional responses that do not include a body.

e.g. the following fails the assertion, but the second response has no `content`
```
```yaml
          responses:
            '200':
              description: Success
              content:
                application/json:
                  schema:
                    $ref: '#/components/schemas/User'
            # Definition of all error statuses
            default:
              description: Unexpected error
```

augustuswm:
I started on an implementation for supporting multiple responses types and came up with a couple options.

Currently each method has some fixed `T` for its `ResponseValue`, and that `T` can be one of:

* A JSON deserializable type
* An empty body, represented as `()`
* A stream of bytes
* A `reqwest::Upgraded` connection

To get to multiple response types a method needs to be able to return a type that may be (at runtime) any of these types. In basic cases where the options are only multiple JSON deserialized types, all of the following are going to be similar for the user of the generated client, and so more of the focus is on problematic endpoints. Also while most of this is focused on the response values, the same patterns can apply to Error responses.

### Composing Types

One option is to represent these as composable types, where `Option<T>` represents a potentially empty body and an upgradable response is represented as:

```rust
// Represents a response that may be upgraded
pub enum Upgradable<T> {
    Upgrade(reqwest::Upgraded),
    Value(T),
}
```

Imagining a method that can return an empty body, an upgraded connection, or a deserialized body (`GetRecordResponse`), it would look like:

```rust
pub async fn get_records<'a>(
  &'a self,
) -> Result<ResponseValue<Upgradable<Option<types::GetRecordsResponse>>>, Error<GetRecordError>> {
  unimplemented!()
}
```

This is a fairly straightforward solution to implement, and is easy (though tedious) to extend with new composing types. I think the experience for an end user of the client though is not great when handling particularly messy endpoints. The optional body case of `ResponseValue<Option<Inner>, Error<Option<InnerError>>>`, is fairly intuitive. But an endpoint with a return value that starts looking like `ResponseValue<ByteStream<Option<Inner>>, Error<Option<InnerError>>>` is pretty verbose and is asking the caller to unpack the full type on every call.

### Progenitor Enum

[...]

Issue #444, opened 2023-04-27
api.github.com


jayvdb:
When I fetch the latest from https://github.com/github/rest-api-description (it has the same version but it is not the same api), I encounter the following five parse errors.  yes, there are only five.

1. fails in typify `schemas_mutually_exclusive` due to this field.

```json
"conditions": {
            "anyOf": [
              {
                "$ref": "#/components/schemas/repository-ruleset-conditions"
              },
              {
                "$ref": "#/components/schemas/org-ruleset-conditions"
              }
            ]
          },
```

2. `/markdown/raw` handles two media types , `text/plain` and `text/x-markdown` .
3. Fails due to this:

```json
"422": {
            "description": "Validation failed",
            "content": {
              "application/json": {
                "schema": {
                  "oneOf": [
                    {
                      "$ref": "#/components/schemas/validation-error"
                    },
                    {
                      "$ref": "#/components/schemas/validation-error-simple"
                    }
                  ]
                }
              }
            }
          },
```

4. Fails due to this

```json
                  "make_latest": {
                    "type": "string",
                    "description": "Specifies whether this release should be set as the latest release for the repository. Drafts and prereleases cannot be set as latest. Defaults to `true` for newly published releases. `legacy` specifies that the latest release should be determined based on the release creation date and higher semantic version.",
                    "enum": [
                      "true",
                      "false",
                      "legacy"
                    ],
                    "default": true
                  },
```

5. `response_types.len() <= 1', progenitor-impl/src/method.rs:1171:9`  https://github.com/oxidecomputer/progenitor/issues/344

jayvdb:
Looks like this overlaps with https://github.com/oxidecomputer/progenitor/issues/395

augustuswm:
I believe all of these are covered in #395, except for maybe the first one. Can you add any details (and the build errors) to the tracking ticket and I'll update the list of issues for building a GitHub client.

PR #445, opened 2023-04-27
Enable github api test

jayvdb:
Relates to https://github.com/oxidecomputer/progenitor/issues/444


progenitor-impl/src/method.rs
```diff
@@ -1168,7 +1168,6 @@ impl Generator {
 
         // TODO to deal with multiple response types, we'll need to create an
         // enum type with variants for each of the response types.
-        assert!(response_types.len() <= 1);
```
jayvdb, line 1171: This fixes https://github.com/oxidecomputer/progenitor/issues/344

I can remove it by adding a lot more fiddling to `test_github`

ahl:
I share this goal, and I share the desire to fail gracefully, but I think that removing an assertion of an expected invariant is not the right direction given the goals of the project.

Issue #395, opened 2023-03-29
Path to GitHub client


augustuswm:
Currently the `octorust` client is generated out of the `third-part-api-clients` repo against an outdated GitHub spec. Ideally progenitor could be used to generate this client against the current live spec instead. This is a tracking issue to track the list of known features needed to make this possible. See the individual tickets for more details.

For reference this list was compiled against: https://github.com/github/rest-api-description/blob/main/descriptions/api.github.com/api.github.com.json

```tasklist
- [ ] Add `text/x-markdown` as a supported request body type
- [ ] Handle enums that specify a default not present in the list of values
- [ ] #344 
- [ ] Handle enum values that contain meaningful non-alphanumeric characters
- [ ] Duplicate struct generation (needs to be validated)
- [ ] Recursive optional type generation (needs to be validated)
- [ ] #268 
- [ ] Support for servers parameter
- [ ] Solution for access / refresh token handling
```

geoffreygarrett:
@augustuswm & @ahl, which of the above tasks relate to the following:

```bash
--- stdout
cargo:rerun-if-changed=./api.github.com.json
--- stderr
thread 'main' panicked at progenitor-github/build.rs:10:50:
called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum ReferenceOr", line: 65608, column: 3)
```
(?)
 
I could only find https://github.com/oxidecomputer/progenitor/issues/348. If desired my setup is as available in the zip in the PR demo https://github.com/oxidecomputer/progenitor/pull/857.

ahl:
> Seems to be the lack of support for JSON schema types in OpenAPI deserialisation.

Yeah, that works in 3.1, but not in 3.0.x. Progenitor *will* support 3.1 at some point, but 1. there aren't that many 3.1 docs out there and 2. I have a stack of work I want to do in order to "propertly" support it

geoffreygarrett:
@ahl I'm not sure if the project is open to PRs etc. or if development will continue, but there's this I was hoping might get a review https://github.com/oxidecomputer/progenitor/pull/857.

geoffreygarrett:
🦗

PR #593, opened 2023-10-05
Add support for text/plain and text/x-markdown body content types

999eagle:
This PR adds support for the `text/plain` and `text/x-markdown` types in request bodies. This should help along with #395.

### DIFF
--- a/progenitor-impl/tests/output/buildomat-builder-tagged.out
+++ b/progenitor-impl/tests/output/buildomat-builder-tagged.out
@@ -1754,6 +1754,18 @@ impl Client {
         builder::Whoami::new(self)
     }
 
+    ///Sends a `PUT` request to `/v1/whoami/name`
+    ///
+    ///```ignore
+    /// let response = client.whoami_put_name()
+    ///    .body(body)
+    ///    .send()
+    ///    .await;
+    /// ```
+    pub fn whoami_put_name(&self) -> builder::WhoamiPutName {
+        builder::WhoamiPutName::new(self)
+    }
+
     ///Sends a `POST` request to `/v1/worker/bootstrap`
     ///
     ///```ignore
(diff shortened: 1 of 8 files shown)


ahl:
Thanks for submitting. Before I review this, can you please include a test and/or an example of the generated output? Happy to advise if you aren't sure how to proceed.

ahl:
I also filed an issue to document how I'd imagine multiple content types working in the fullness of time (to be clear, not something I'd expect in this PR)

https://github.com/oxidecomputer/progenitor/issues/594

999eagle:
<details><summary>Example for the generated output (this is taken from the Keycloak API which takes some massaging to actually generate code)</summary>
<pre>
      "put" : {
        "tags" : [ "Realms Admin" ],
        "parameters" : [ {
          "name" : "key",
          "in" : "path",
          "required" : true,
          "schema" : {
            "type" : "string"
          }
        }, {
          "name" : "locale",
          "in" : "path",
          "required" : true,
          "schema" : {
            "type" : "string"
          }
        } ],
        "requestBody" : {
          "content" : {
            "text/plain" : {
              "schema" : {
                "type" : "string"
              }
            }
          }
        },
        "responses" : {
          "204" : {
            "description" : "No Content"
          }
        }
      },
</pre>
<pre>
    pub async fn put_realm_localization_locale_key<'a>(
        &'a self,
        realm: &'a str,
        locale: &'a str,
        key: &'a str,
        body: String,
    ) -> Result<ResponseValue<()>, Error<()>> {
        let url = format!(
            "{}/admin/realms/{}/localization/{}/{}", self.baseurl, encode_path(& realm
            .to_string()), encode_path(& locale.to_string()), encode_path(& key
            .to_string()),
        );
        let request = self
            .client
            .put(url)
            .header(
                reqwest::header::CONTENT_TYPE,
                reqwest::header::HeaderValue::from_static("text/plain"),
            )
            .body(body)
            .build()?;
        let result = self.client.execute(request).await;
        let response = result?;
        match response.status().as_u16() {
            204u16 => Ok(ResponseValue::empty(response)),
            _ => Err(Error::UnexpectedResponse(response)),
        }
    }
</pre>
</details>

I'd appreciate if you could help me make this into a more minimal test.

ahl:
I often throw something in here when I want to add a simple system test: https://github.com/oxidecomputer/progenitor/blob/main/sample_openapi/buildomat.json

999eagle:
I've made up an operation to add to the `buildomat.json` spec and updated the expected test outputs to test this new feature. Only cli support is missing from this PR as the cli currently doesn't have support for non json encoded bodies anyway.

ahl (commented):
Thanks for adding the test; much easier to see what code this generates. I left one note where I think the generation could use some work. Let me know if you disagree or think that the advance in utility is already sufficient.

999eagle:
I don't see any review notes on GitHub, could you please post them again?

progenitor-impl/tests/output/buildomat-builder-tagged.out
```diff
@@ -2355,6 +2367,51 @@ pub mod builder {
...
+
+    impl<'a> WhoamiPutName<'a> {
+        pub fn new(client: &'a super::Client) -> Self {
+            Self {
+                client,
+                body: Err("body was not initialized".to_string()),
+            }
+        }
+
+        pub fn body(mut self, value: String) -> Self {
```
ahl, line 2387: In other situations where there's a string parameter we have code like this:
```rust
        pub fn task<V>(mut self, value: V) -> Self
        where
            V: std::convert::TryInto<String>,
        {
```

Should we do something similar here rather than just an explicit `String`?

ahl:
> I don't see any review notes on GitHub, could you please post them again?

Sorry! My mistake. Updated now.

progenitor-impl/tests/output/buildomat-builder-tagged.out
999eagle: Yeah, that sounds reasonable! I've changed the builder implementation now to look like this.

ahl:
Thanks for this; where are you using it?

999eagle:
I'm using progenitor to generate an API client for the [Keycloak](https://github.com/keycloak/keycloak) admin API.

ahl:
Awesome! Is it working now? Thanks again! Have you considered making a crate on crates.io?

999eagle:
It's mostly working now, yes. I still have to work around some things like missing operation ids, array parameters and bodies with multiple allowed mime types, but with the right build script it works for what I need. I might publish a crate for this later when I get the high level wrappers into a more useful shape.

Issue #288, opened 2022-12-28
`field already declared` when using the Builder interface with a parameter called: `client`


dmeijboom:
Using the following code:

```rust
generate_api!(interface = Builder, spec = "filename.json");
```

And the following spec:

```json
{
    "openapi": "3.0.0",
    "info": {
        "title": "API",
        "version": "0.0.1"
    },
    "servers": [],
    "components": {
        "schemas": {
            "TimeEntry": {
                "type": "object",
                "required": [],
                "properties": {
                    "client": {
                        "type": "string"
                    }
                }
            }
        }
    },
    "paths": {
        "/time_entries": {
            "get": {
                "summary": "Returns a list of time entries",
                "operationId": "listTimeEntries",
                "parameters": [
                    {
                        "name": "client",
                        "in": "query",
                        "description": "Client",
                        "required": false,
                        "schema": {
                            "type": "string"
                        }
                    },
                ],
                "responses": {
                    "200": {
                        "description": "A JSON array of time-entries",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "array",
                                    "items": {
                                        "$ref": "#/components/schemas/TimeEntry"
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

Results in the following code being generated for the `TimeEntry` component:

```rust
    pub struct ListTimeEntries<'a> {
        client: &'a super::Client,
        client: Result<Option<String>, String>,
    }
```

This breaks with the following exception:

```shell
➜ cargo check
    Checking [redacted] v0.1.0 (/[redacted])
error[E0416]: identifier `client` is bound more than once in the same pattern
 --> src/[redacted].rs:8:1
  |
8 | generate_api!(interface = Builder, spec = "filename.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ used in a pattern more than once

error[E0124]: field `client` is already declared
 --> src/[redacted].rs:8:1
  |
8 | generate_api!(interface = Builder, spec = "filename.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | |
  | field already declared
  | `client` first declared here
  |
[...]

Issue #374, opened 2023-03-21
Parameter named `url` omitted due to conflicting name in the code


provokateurin:
I have an operation with a query parameter named `url`. The generated code looks like this:
```rust
    #[doc = "Sends a `POST` request to `/index.php/apps/news/api/v1-3/feeds`\n\n"]
    pub async fn news_feed_api_add_feed<'a>(
        &'a self,
        folder_id: Option<i64>,
        url: &'a str,
    ) -> Result<ResponseValue<types::NewsListFeeds>, Error<()>> {
        let url = format!("{}/index.php/apps/news/api/v1-3/feeds", self.baseurl,);
        let mut query = Vec::with_capacity(2usize);
        if let Some(v) = &folder_id {
            query.push(("folderId", v.to_string()));
        }

        query.push(("url", url.to_string()));
        let request = self.client.post(url).query(&query).build()?;
        let result = self.client.execute(request).await;
        let response = result?;
        match response.status().as_u16() {
            200u16 => ResponseValue::from_response(response).await,
            _ => Err(Error::UnexpectedResponse(response)),
        }
    }
```
The problem is that the code defines the variable `url` again, so the parameter is never used.

provokateurin:
I think inlining the variable would do the trick, but the same could happen with a parameter named `query`.

ahl:
Yeah this is an unfortunate problem that a few folks have hit; I'll get this sorted
See #261

inickles:
This has been fixed with https://github.com/oxidecomputer/progenitor/pull/599.

PR #599, opened 2023-10-11
Prefix internal variable names to help avoid collisions

inickles:
This PR is essentially a reboot of https://github.com/oxidecomputer/progenitor/pull/261, using a name prefix for internal variables as discussed in that PR.

I saw the discussion of possible fixes in I ran into https://github.com/oxidecomputer/progenitor/issues/288. I decided to go with the internal name prefix approach because it seems to be the simplest. While not exactly 100% perfect and checking for collisions ahead of time probably would be, I think this PR is gets close enough to make it worthwhile.

There is a slight change to the `compile` function in `template.rs` which was done to bring the reference to the url variable back into the calling function so we wouldn't have to export the formatted identity variable, or define it in both locations (yuck).

Before:
```rust
        let url = format!(
            "{}/admin/directory/v1/customer/{}/devices/chromeos", self.baseurl,
            encode_path(& customer_id.to_string()),
        );
        let mut query = Vec::with_capacity(19usize);
        ...
        if let Some(v) = &query {
            query.push(("query", v.to_string()));
        }
        ...
        let request = self
            .client
            .get(url)
            .header(
                reqwest::header::ACCEPT,
                reqwest::header::HeaderValue::from_static("application/json"),
            )
            .query(&query)
            .build()?;
        let result = self.client.execute(request).await;
        let response = result?;
        match response.status().as_u16() {
            200u16 => ResponseValue::from_response(response).await,
            _ => Err(Error::UnexpectedResponse(response)),
        }
```

After:
```rust
        let __progenitor_url = format!(
            "{}/admin/directory/v1/customer/{}/devices/chromeos",
            self.baseurl,
            encode_path(&customer_id.to_string()),
        );
        ...
        if let Some(v) = &query {
            __progenitor_query.push(("query", v.to_string()));
        }
        ...
        let __progenitor_request = self
            .client
            .get(__progenitor_url)
            .header(
                reqwest::header::ACCEPT,
                reqwest::header::HeaderValue::from_static("application/json"),
            )
            .query(&__progenitor_query)
            .build()?;
        let __progenitor_result = self.client.execute(__progenitor_request).await;
        let __progenitor_response = __progenitor_result?;
        match __progenitor_response.status().as_u16() {
            200u16 => ResponseValue::from_response(__progenitor_response).await,
            _ => Err(Error::UnexpectedResponse(__progenitor_response)),
        }
```

### DIFF
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -792,6 +792,16 @@ impl Generator {
         method: &OperationMethod,
         client: TokenStream,
     ) -> Result<MethodSigBody> {
+        // We prefix internal variable names to mitigate possible name
+        // collisions with the input spec. See:
+        // https://github.com/oxidecomputer/progenitor/issues/288
+        let internal_prefix = "__progenitor";
+        let url_var = format_ident!("{internal_prefix}_url");
+        let query_var = format_ident!("{internal_prefix}_query");
+        let request_var = format_ident!("{internal_prefix}_request");
+        let response_var = format_ident!("{internal_prefix}_response");
+        let result_var = format_ident!("{internal_prefix}_result");
+
         // Generate code for query parameters.
         let query_items = method
             .params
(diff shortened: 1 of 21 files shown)


ahl:
Yeah fine. I think we need to update the test "fixtures"

inickles:
> Yeah fine. I think we need to update the test "fixtures"

This is now done and tests are passing.

ahl (approved):
cool; thanks

ahl:
@inickles are there bugs we can now close as a result of this?

inickles:
I closed an issue and a PR. Not sure if this covers it, but I gave the open issues a quick scan.

I'd originally commented that https://github.com/oxidecomputer/progenitor/issues/288 is the issue i ran into, but that wasn't correct, I updated the comment to say https://github.com/oxidecomputer/progenitor/issues/374, which is.

https://github.com/oxidecomputer/progenitor/issues/288 is similar and I think is worthy of a new PR.

PR #601, opened 2023-10-13
Generate unique internal var names and add tests for name collisions.

inickles:
Ok we're back, one last time for now. 🤞

There are two changes here:

1. @augustuswm suggested that I add the test he'd introduced in the now-closed https://github.com/oxidecomputer/progenitor/pull/261.
2. In doing so I realized generating a unique name for internal variables from the params wouldn't be hard and it'd create less noise in generated output. I feel better about this than the static prefixes I introduces earlier.

Example output from new test:
```rust
impl Client {
    ///Gets a key
    ///
    ///Sends a `GET` request to `/key/{query}`
    ///
    ///Arguments:
    /// - `query`: Parameter name that was previously colliding
    /// - `client`: Parameter name that was previously colliding
    /// - `request`: Parameter name that was previously colliding
    /// - `response`: Parameter name that was previously colliding
    /// - `result`: Parameter name that was previously colliding
    /// - `url`: Parameter name that was previously colliding
    pub async fn key_get<'a>(
        &'a self,
        query: bool,
        client: bool,
        request: bool,
        response: bool,
        result: bool,
        url: bool,
    ) -> Result<ResponseValue<()>, Error<()>> {
        let _url = format!("{}/key/{}", self.baseurl, encode_path(&query.to_string()),);
        let mut _query = Vec::with_capacity(5usize);
        _query.push(("client", client.to_string()));
        _query.push(("request", request.to_string()));
        _query.push(("response", response.to_string()));
        _query.push(("result", result.to_string()));
        _query.push(("url", url.to_string()));
        let _request = self.client.get(_url).query(&_query).build()?;
        let _result = self.client.execute(_request).await;
        let _response = _result?;
        match _response.status().as_u16() {
            200u16 => Ok(ResponseValue::empty(_response)),
            _ => Err(Error::UnexpectedResponse(_response)),
        }
    }
}
```

### DIFF
--- a/progenitor-impl/src/util.rs
+++ b/progenitor-impl/src/util.rs
@@ -123,3 +123,21 @@ pub(crate) fn sanitize(input: &str, case: Case) -> String {
         format!("{}_", out)
     }
 }
+
+/// Given a desired name and a slice of proc_macro2::Ident, generate a new
+/// Ident that is unique from the slice.
+pub(crate) fn unique_ident_from(
+    name: &str,
+    identities: &[proc_macro2::Ident],
+) -> proc_macro2::Ident {
+    let mut name = name.to_string();
+
+    loop {
+        name.insert_str(0, "_");
+        let ident = quote::format_ident!("{}", name);
+
+        if !identities.contains(&ident) {
+            return ident;
+        }
+    }
+}
--- a/progenitor-impl/src/util.rs
+++ b/progenitor-impl/src/util.rs
@@ -133,11 +133,12 @@ pub(crate) fn unique_ident_from(
     let mut name = name.to_string();
 
     loop {
-        name.insert_str(0, "_");
         let ident = quote::format_ident!("{}", name);
 
         if !identities.contains(&ident) {
             return ident;
         }
+
+        name.insert_str(0, "_");
     }
 }
--- a/sample_openapi/param-collision.json
+++ b/sample_openapi/param-collision.json
@@ -1,0 +1,78 @@
+{
+    "openapi": "3.0.0",
+    "info": {
+        "description": "Minimal API for testing collision between parameter names and generated code",
+        "title": "Parameter name collision test",
+        "version": "v1"
+    },
+    "paths": {
+        "/key/{query}": {
+            "get": {
+                "description": "Gets a key",
+                "operationId": "key.get",
+                "parameters": [
+                    {
+                        "description": "Parameter name that was previously colliding",
+                        "in": "path",
+                        "name": "query",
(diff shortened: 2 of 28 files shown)


augustuswm:
One small suggestion. Does it make sense here to use an underscore suffix instead of a prefix due to the underscore prefix having a semantic meaning of "may be unused" already? Given that it is all occurring in generated code, I don't think there is a worry of correctness. As a reader of the generated code though, you may wonder why they are all marked that way.

inickles:
> One small suggestion. Does it make sense here to use an underscore suffix instead of a prefix due to the underscore prefix having a semantic meaning of "may be unused" already? Given that it is all occurring in generated code, I don't think there is a worry of correctness. As a reader of the generated code though, you may wonder why they are all marked that way.

Yeah, I could see that. Though my general preference for this is for prefixing, the differentiation is clearer to me, so if we wanted to alleviate that potential confusion my preference would be to use a double underscore, which I've seen used for private things.

I could change the string in https://github.com/oxidecomputer/progenitor/pull/601/files#diff-25893ecdd4f4d0e002500c89a463ff6bc8e30149732f3cb90dd1bc247c5e788cR136 to just be a `"__"`, which would increase the prefix by two at a time, still always prefixing by default.

I don't care if we choose to increment by one or two underscores (or perhaps an emoji instead) upon collision, or even if we default to using a prefix from the start - I could go either way on both those points and I'll defer to you @augustuswm  and @ahl. Simplest option seems to be just replace the str with `"__"`.

progenitor-impl/src/util.rs
```diff
@@ -123,3 +123,21 @@ pub(crate) fn sanitize(input: &str, case: Case) -> String {
...
+    let mut name = name.to_string();
+
+    loop {
+        name.insert_str(0, "_");
+        let ident = quote::format_ident!("{}", name);
+
+        if !identities.contains(&ident) {
+            return ident;
+        }
+    }
```
ahl, line 142: why not:
```suggestion
    loop {
        let ident = quote::format_ident!("{}", name);

        if !identities.contains(&ident) {
            return ident;
        }
        
        name.insert_str(0, "_");
    }
```

sample_openapi/param-collision.json
```diff
@@ -0,0 +1,78 @@
+{
```
ahl, line 1: where is this file used?

inickles: oops - somehow missed the important part where we invoke a test that uses it.. that's fixed now.

ahl:
nice

PR #606, opened 2023-10-16
Add CLI alias for standard in

augustuswm:
* Adds `-` as an alias for standard in when specifying a json body

This mirrors how the `api` method of the oxide.rs CLI functions. Body handling now looks like (for some type T):

```rust
if let Some(value) = matches.get_one::<String>("json-body") {
    use std::io::Read;
    let body_input = match value.as_str() {
        "-" => {
            let mut buf = Vec::new();
            std::io::stdin().read_to_end(&mut buf).unwrap();
            buf
        }
        file => std::fs::read(&file).unwrap(),
    };
    let body_value =
        serde_json::from_slice::<T>(&body_input).unwrap();
    request = request.body(body_value);
}
```

### DIFF
--- a/progenitor-impl/src/cli.rs
+++ b/progenitor-impl/src/cli.rs
@@ -440,7 +440,7 @@ impl Generator {
             CliBodyArg::Optional => Some(false),
         })
         .map(|required| {
-            let help = "Path to a file that contains the full json body.";
+            let help = r#"Path to a file that contains the full json body (use "-" to read from standard input)."#;
 
             quote! {
                 .arg(
@@ -450,7 +450,7 @@ impl Generator {
                         // Required if we can't turn the body into individual
                         // parameters.
                         .required(#required)
-                        .value_parser(clap::value_parser!(std::path::PathBuf))
+                        .value_parser(clap::value_parser!(String))
                         .help(#help)
                 )
                 .arg(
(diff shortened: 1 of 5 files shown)


progenitor-impl/src/cli.rs
```diff
@@ -477,12 +477,21 @@ impl Generator {
...
-                    let body_txt = std::fs::read_to_string(value).unwrap();
+                    use std::io::Read;
+                    let body_input = match value.as_str() {
+                        "-" => {
+                            let mut buf = Vec::new();
+                            std::io::stdin().read_to_end(&mut buf).unwrap();
+                            buf
+                        }
+                        file => std::fs::read(&file).unwrap()
+                    };
```
ahl, line 490: Would `from_reader` simplify this at all? https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html

augustuswm: Definitely, updated it to use `from_reader` in both branches. Before merging though I would like to see if I can get a test in here. I don't think anything currently tests if a generated CLI would actually compile.

ahl: That would be great

ahl:
@augustuswm as I recall you were going to add a test; no rush, just wanted to make sure you weren't blocked on me.

augustuswm:
Correct, not blocked. On me to get that added.

augustuswm:
Added a rough pass at a compilation test. It is pretty much just a macro to generates tokens for each sdk and cli. These are then wrapped in tests that will fail to compile if the generated code fails to compile. The rest of the changes are largely bringing in the dev dependencies that the generated sdks and clis need to compile.

Issue #402, opened 2023-04-03
form-data as content type yet to be implemented


drahnr:
```
    post:
      tags:
      - "Voucher"
      summary: "Free shit"
      description: "No money needed!"
      operationId: "uploadVoucher"
      requestBody:
        description: "File to upload"
        content:
          form-data:
            schema:
              properties:
                file:
                  description: "The file to upload"
                  type: "string"
                  format: "binary"
              type: "object"
```

```
  Error: UnexpectedFormat("unexpected content type: form-data")
```

drahnr:
Mea culpa, must be `"application/x-www-form-urlencoded":` instead of `form-data`, assuming upstream change.

Thank you for this beautiful tool :heart_decoration:

drahnr:
So while the above statement is correct, the behavior should be different. See https://github.com/oxidecomputer/progenitor/issues/410

Issue #177, opened 2022-08-26
GC unused types


ahl:
Typify can generate types that are unreachable from any progenitor interface. How? In some cases, typify can get pretty fancy and infer useful structure from referenced types. For example, there may be named, referenced types that are used as the variants of a `oneOf`. In that case, typify may generate both the referenced types and, effectively, partial copies of those types embedded within an `enum`.

Only types directly or transitively referenced by operations actually need to be in the final code.

jayvdb:
If this is done, it would be good to have an option to include all schemas that are literally in the input document.  I am using progenitor to just generate the models/types, which are used by other custom crates.

ahl:
@jayvdb would the typify crate work for you?

jayvdb:
We are using the progenitor CLI, with lots of customisations in typify and in progenitor.  If you force removal of unused OpenAPI schemas, I will need to add another hack to progenitor to disable that behaviour.  Not too hard, but I am hoping to eventually be using the progenitor crate.

ahl:
This would be optional. Not required.

PR #609, opened 2023-10-22
Support multipart/form-data and file uploads

### LINKED ISSUE #518: content multipart/form-data is not supported
```
UnexpectedFormat("unexpected content type: multipart/form-data")
```

```yaml
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                image_file:
                  type: string
                  format: binary
```

https://swagger.io/docs/specification/describing-request-body/multipart-requests/

ahirner:
Fixes #518 and #402.

This approach factors out binary form data as `reqwest::multipart::Part` parameter. This has the following advantages:
- Clients gains the ability to stream large files
- Bodies remain `serde::Serialize` which they couldn't if a `Part` was made from an async stream

The disadvantage is leaking a `reqwest` type. However, we could probably wrap and imple `Into` for an internal type if required.

An internal media heavy API compiled well. I'm looking for guidance and tips. Todos:
- [x] decide and finalize how to `typify` binary formatted strings
- [x] some work on cli
- [x] some work on httpmock
- [x] tests


ahl:
I didn't take a close look at this as it appears to still be a work in progress, but I did want to say:

1. Thank you! great to see folks participating in this project
2. I've found that one of the best ways to demonstrate new features such as this one is to add an end-to-end test by adding an operation in an existing OpenAPI json file or by adding a new file; this will make the output clear in the updated fixture data.

progenitor-client/src/progenitor_client.rs
```diff
@@ -416,3 +416,17 @@ impl<E> RequestBuilderExt<E> for RequestBuilder {
             })?))
     }
 }
+
+#[doc(hidden)]
```
jayvdb, line 420: imo this should be documented

progenitor-impl/src/method.rs
```diff
@@ -2224,16 +2331,223 @@ impl Generator {
...
+                            let v = to_form_string(v)?;
+                            #form_ident = #form_ident.text(#prop_name, v);
+                        };
+                    }
+                });
+            form_parts_optional.extend(body_parts);
+        }
+
+        // only build if any param or at least a body without additional parts
+        if form_parts.len() > 0
```
jayvdb, line 2472: `!form_parts.is_empty()` here, and on next line, and `let mutt` a few lines lower.

ahirner:
> I've found that one of the best ways to demonstrate new features such as this one is to add an end-to-end test

Check out `progenitor-impl/tests/output/openai-openapi-*.out`. This API features multiple files in one operation and optionals as well.

For an integrated example see: https://github.com/ahirner/openai-progenitor

> decide and finalize how to typify 

The outputs are fully functional AFAIC. My last gripe is that `multipart/form-data` body structs are emitted twice. Once with file fields factored out and once from the original `ref`. My idea would be to inline those definitions before adding them: https://github.com/oxidecomputer/progenitor/blob/a6bc0624803b9483983f03269533cc35db66a54b/progenitor-impl/src/lib.rs#L223

LMK what you think!
Btw, I think there is also the opportunity to DRY the token generating methods.

progenitor-impl/src/method.rs
```diff
@@ -1640,6 +1714,7 @@ impl Generator {
                             // a `body_map()` method that operates on the
                             // builder itself.
                             (Some(builder_name), false) => {
+                                // todo: rm for ...Kind::FormData
```
jayvdb, line 1717: just noting there is a todo here

progenitor-impl/src/method.rs
```diff
@@ -138,6 +142,7 @@ pub enum BodyContentType {
     Json,
     FormUrlencoded,
     Text(String),
+    FormData(bool),
```
jayvdb, line 145: would be good to add doc here to indicate that the `bool` (iiuc) holds whether it is required or not.  It feels like this is a strange spot to store that info.

progenitor-impl/src/method.rs
```diff
@@ -438,8 +445,16 @@ impl Generator {
             self.uses_websockets = true;
         }
 
-        if let Some(body_param) = self.get_body_param(operation, components)? {
-            params.push(body_param);
+        params.extend(self.get_body_params(operation, components)?);
+        // check if some bodies use serde_json and more for serialization
```
jayvdb, line 449: this comment is a bit hard to understand

progenitor-impl/src/method.rs
```diff
@@ -2224,16 +2331,223 @@ impl Generator {
                     .add_type_with_name(&schema.to_schema(), Some(name))?;
                 OperationParameterType::Type(typ)
             }
+            BodyContentType::FormData(_) => {
+                // remove FormPart properties
+                // TODO: move into util
```
jayvdb, line 2336: another todo

progenitor-impl/src/method.rs
```diff
@@ -2224,16 +2331,223 @@ impl Generator {
...
+                        object: Some(ref mut object),
+                        ..
+                    },
+                ) = schema
+                {
+                    for param in body_params.iter() {
+                        object.properties.remove_entry(&param.api_name);
+                    }
+                };
+                // TODO: because add_ref_types is called before, the unmodified
```
jayvdb, line 2365: another todo

progenitor-impl/src/method.rs
```diff
@@ -2224,16 +2331,223 @@ impl Generator {
...
+        .iter()
+        .filter_map(|(key, item)| match item {
+            ReferenceOr::Item(i) => Some((key, i)),
+            _ => None,
+        })
+        .filter_map(|(key, i)| {
+            match (&i.schema_data, &i.schema_kind) {
+                (
+                    openapiv3::SchemaData {
+                        // todo: assert nullable if not required?
```
jayvdb, line 2514: another todo

progenitor-impl/src/method.rs
```diff
@@ -633,6 +648,12 @@ impl Generator {
                         }
                     }
                     (OperationParameterType::RawBody, true) => unreachable!(),
+                    (OperationParameterType::FormPart, false) => {
```
jayvdb, line 651: move these above the `(OperationParameterType::RawBody,..` case which is entirely `unreachable!()` , as it is nice to have invalid cases at the bottom

progenitor-client/src/progenitor_client.rs
ahirner: Added docstring. I think hiding shrinks the public interface, while client code can reside in a different crate than the generated client lib.

progenitor-impl/src/method.rs
ahirner: Documented it. 
It's the least surprising spot, since it follows the convention of parent variants in `OperationParameterKind` having such a member. That is, until general optionality of body parameters is solved (not in this PR).

progenitor-impl/src/lib.rs
```diff
@@ -60,6 +61,43 @@ pub struct GenerationSettings {
...
+        let bin_string = SchemaObject {
+            instance_type: Some(schemars::schema::SingleOrVec::Single(
+                Box::new(InstanceType::String),
+            )),
+            format: Some("binary".to_string()),
+            ..SchemaObject::default()
+        };
+        // This treats binary strings on the level of generated types.
+        // For now, we'd want to remove this field because
+        // the only supported binary string is handled as FormPart instead.
```
jayvdb, line 77: "the only supported binary string" - maybe this needs rewording, as progenitor does support:
```yaml
      requestBody:
        content:
          application/octet-stream:
            schema:
              type: string
              format: binary
```
c.f. https://github.com/oxidecomputer/progenitor/blob/a6bc0624803b9483983f03269533cc35db66a54b/progenitor-impl/src/method.rs#L149

ahirner:
PR is complete and all comments so far adressed.

ahirner:
Hi @ahl, is there maybe something I can do in the meantime?

progenitor-client/src/progenitor_client.rs
```diff
@@ -416,3 +416,32 @@ impl<E> RequestBuilderExt<E> for RequestBuilder {
...
+/// ```
+#[doc(hidden)]
+pub fn to_form_string<E>(
+    value: &serde_json::Value,
+) -> Result<String, Error<E>> {
+    match value.as_str() {
+        // Take unquoted String Value
+        Some(v) => Ok(v.to_string()),
+        // otherwise convert to quoted json
+        None => serde_json::to_string(value).map_err(|_| {
```
ahl, line 443: doesn't this make `serde_json` into a mandatory dependency for generated crates?

progenitor-impl/src/method.rs
```diff
@@ -2224,18 +2337,207 @@ impl Generator {
...
+                } else {
+                    form_parts_optional.push(quote! {
+                        if let Some(#param_ident) = #param_ident {
+                            #form_ident = #form_ident.part(#api_name, #param_ident);
+                        };
+                    });
+                };
+            });
+
+        // set params from form
```
ahl, line 2423: Sorry. I'm not sure what this comment is intended to convey.

progenitor-impl/src/method.rs
```diff
@@ -2224,18 +2337,207 @@ impl Generator {
...
+                if required {
+                    form_parts.push(quote! { .part(#api_name, #param_ident) });
+                } else {
+                    form_parts_optional.push(quote! {
+                        if let Some(#param_ident) = #param_ident {
+                            #form_ident = #form_ident.part(#api_name, #param_ident);
+                        };
+                    });
+                };
+            });
```
ahl, line 2421: Let's get rid of the filter_map with side-effects. Let's just do `for param in params { ` and go from there.

progenitor-impl/src/method.rs
```diff
@@ -2224,18 +2337,207 @@ impl Generator {
...
+        body_params.push(body_param);
+
+        Ok(body_params)
+    }
+
+    fn get_form_builder(
+        &self,
+        params: &[OperationParameter],
+        form_ident: &proc_macro2::Ident,
+    ) -> Result<Vec<TokenStream>> {
```
ahl, line 2383: each item of the response Vec should represent a similar sort of thing. Currently it looks like kind of a mishmash. I think this would be clearer if we could make each item reflect a similar type of thing and document the concept this function is implementing.

progenitor-impl/src/lib.rs
```diff
@@ -308,15 +337,24 @@ impl Generator {
         };
 
         let version_str = &spec.info.version;
+        let (to_form_string, pub_part) = if self.uses_form_parts {
+            (
+                Some(quote! { ,to_form_string }),
+                Some(quote! {pub use reqwest::multipart::Part; }),
```
ahl, line 343: let's just fully specify the path everywhere this is used please.

progenitor-impl/src/lib.rs
```diff
@@ -308,15 +337,24 @@ impl Generator {
         };
 
         let version_str = &spec.info.version;
+        let (to_form_string, pub_part) = if self.uses_form_parts {
+            (
+                Some(quote! { ,to_form_string }),
```
ahl, line 342: use it unconditionally (note #[allow(unused_imports)])

progenitor-impl/src/lib.rs
```diff
@@ -207,20 +209,47 @@ impl Generator {
...
+            .paths
+            .iter()
+            .flat_map(|(_path, ref_or_item)| {
+                // Exclude externally defined path items.
+                let item = ref_or_item.as_item().unwrap();
+                item.iter().map(move |(_method, operation)| operation)
+            })
+            .flat_map(|operation| operation.request_body.iter())
+            .filter_map(|b| b.as_item().map(|b| &b.content))
+            .flat_map(|body_content| body_content.iter());
```
ahl, line 231: we're trying to exclude these so we don't generate types that aren't used?

progenitor-impl/src/lib.rs
```diff
@@ -207,20 +209,47 @@ impl Generator {
...
         }
     }
 
-    pub fn generate_tokens(&mut self, spec: &OpenAPI) -> Result<TokenStream> {
+    fn add_ref_types(&mut self, spec: &OpenAPI) -> Result<()> {
         validate_openapi(spec)?;
 
+        // collect all component names in request body media references
+        // that are only used for multipart/form-data operations.
+        let media_conents = spec
```
ahl, line 221: typo

progenitor-impl/src/lib.rs
```diff
@@ -207,20 +209,47 @@ impl Generator {
...
+            .paths
+            .iter()
+            .flat_map(|(_path, ref_or_item)| {
+                // Exclude externally defined path items.
+                let item = ref_or_item.as_item().unwrap();
```
ahl, line 226: why is this valid?

progenitor-impl/src/lib.rs
```diff
@@ -581,10 +630,51 @@ pub fn validate_openapi(spec: &OpenAPI) -> Result<()> {
     Ok(())
 }
 
+/// Return all component names where schema references are keyed
+/// by multipart/form-data but nowhere else. The last part of a media
+/// schema's json path reference is considered the component name.
+fn get_exclusive_formdata_refs<'a>(
```
ahl, line 636: these checks seem insufficient to validate if referenced types are used elsewhere e.g. what if the type is used as a body parameter or response parameter. I may be misunderstanding what's going on here, but I don't think we should do any of this.

ahl (changes requested):
This is not exhaustive, but what I could get through. I'd encourage you to generalize from these comments and go through the PR yourself again. Thank you for the submission and my apologies for the delay reviewing it.

progenitor-impl/src/lib.rs
ahirner: > what if the type is used as a body parameter or response parameter

That's what this function should find out. Alas, it's only partially validated in tests, not provably.

> what's going on here

It's to avoid that multipart/form-data body structs are emitted twice. Once with file fields factored out and once from the original ref.

> should do any of this

I thinkt so too, but haven't found a cooler way. My initial idea was way to inline types earlier, but I think the current typify API didn't let me. Maybe I'll find a better approach still.

ahl: I would suggest we just not do anything clever here: generating orphaned types is fine--we already do it!

ahirner: Ok... good to know! Then this PR might come to a conclusion. 
It's only that without https://github.com/oxidecomputer/progenitor/pull/609/commits/f8baafaacc7d215f5c7402c4d7bce5b33c6d265f, the number of orphaned types was very, very large for our internal uses. This simply went against my sense of aesthetics. Technically, I don't think it added compile times really.

ahl: there's probably a bigger, more holistic look one might take: start from all operations, see the types they touch. We could have special handling for these multi-part types (i.e. use in those cases wouldn't count as real use)

ahirner: Yes. OTOH, the more I thought about that, the larger the PR grew in my mind as well ;)
[Intuitively](https://www.reddit.com/r/rust/comments/17pyoo5/5_hours_to_77_seconds_how_database_tricks_sped_up/k8brg3r/), every operation could probably inferred from a generalized query into a graph of schema units.

ahirner:
Obviously this PR stalled due to my lack of effort.
Overall I suggest to revisit it after #384. #177 would make the generated code leaner also.

PR #614, opened 2023-10-29
Require docs for public members

jayvdb:


progenitor-client/src/progenitor_client.rs
```diff
@@ -386,7 +386,7 @@ const PATH_SET: &percent_encoding::AsciiSet = &percent_encoding::CONTROLS
     .add(b'/')
     .add(b'%');
 
-#[doc(hidden)]
+/// Percent encode input string.
```
ahl, line 389: why doc this? this is really only for internal use.

progenitor-client/src/lib.rs
```diff
@@ -1,5 +1,9 @@
 // Copyright 2022 Oxide Computer Company
 
+//! Client support for the progenitor OpenAPI client generator.
```
ahl, line 3: What do you think?

```suggestion
//! Support for generated clients.
```

progenitor-client/src/progenitor_client.rs
jayvdb: It is part of the public API.

ahl: Ok: I don't want to document this. I don't want people using it explicitly.

jayvdb: ok, ill revert this bit

ahl (approved):

jayvdb:
@ahl , PR refreshed & green.

PR #635, opened 2023-11-26
Avoid generating needless lifetimes

jayvdb:


ahl:
I think there are still cases where we generate lifetimes where we don't need it. My preference would be to either avoid them entirely or just generate them all the time. The benefit of avoiding needless lifetimes is that people wouldn't need to ignore the lint; the benefit of generating them all the time is simplicity of codegen. I think that not generating them sometimes means that people (may) still need to ignore the lint *and* the code is more complex.

jayvdb:
This avoids all of the needless lifetimes in my generated clients, covering 51 endpoints.

I'll dig into how to prove there are no other needless lifetimes. :/

jayvdb:
fyi, you were right.  https://github.com/oxidecomputer/progenitor/pull/638 allowed me to see some other cases.  I'll see if they can be removed without adding too much complexity.

ahl:
Sounds good. I think it may be that the complexity of avoiding needless lifetimes may outweigh the benefit... but I'd love to be proved wrong!

Issue #637, opened 2023-12-03
Generated httpmock Then.ok & .json_body_obj compile issues


jayvdb:
When I attempt to compile the generated output, I see

```rust
error[E0308]: mismatched types
   --> src/propolis_server_httpmock.rs:302:39
    |
302 |             Self(self.0.json_body_obj(value))
    |                         ------------- ^^^^^ expected `&_`, found `InstanceStateRequested`
    |                         |
    |                         arguments to this method are incorrect
    |
    = note: expected reference `&_`
                    found enum `propolis_server_builder::types::InstanceStateRequested`
note: method defined here
   --> /home/jayvdb/.cargo/registry/src/index.crates.io-6f17d22bba15001f/httpmock-0.6.8/src/api/spec.rs:534:12
    |
534 |     pub fn json_body_obj<'a, T>(self, body: &T) -> Self
    |            ^^^^^^^^^^^^^
help: consider borrowing here
    |
302 |             Self(self.0.json_body_obj(&value))
    |                                       +

Some errors have detailed explanations: E0308, E0433.
For more information about an error, try `rustc --explain E0308`.
```

when I add `&` as suggested, I get errors like

```rust
error[E0277]: the trait bound `&propolis_server_builder::types::InstanceStateMonitorRequest: Deserialize<'_>` is not satisfied
   --> src/propolis_server_httpmock.rs:356:39
    |
356 |             Self(self.0.json_body_obj(&value))
    |                         ------------- ^^^^^^ the trait `Deserialize<'_>` is not implemented for `&propolis_server_builder::types::InstanceStateMonitorRequest`
    |                         |
    |                         required by a bound introduced by this call
    |
note: required by a bound in `When::json_body_obj`
   --> /home/jayvdb/.cargo/registry/src/index.crates.io-6f17d22bba15001f/httpmock-0.6.8/src/api/spec.rs:536:24
    |
534 |     pub fn json_body_obj<'a, T>(self, body: &T) -> Self
    |            ------------- required by a bound in this associated function
535 |     where
536 |         T: Serialize + Deserialize<'a>,
    |                        ^^^^^^^^^^^^^^^ required by this bound in `When::json_body_obj`
help: consider removing the leading `&`-reference
    |
356 -             Self(self.0.json_body_obj(&value))
356 +             Self(self.0.json_body_obj(value))
    |

```

[...]

PR #638, opened 2023-12-03
Move test generated clients into a buildable crate

jayvdb:
This only moves them into a crate to allow building them.
The next step would be to add a CI job that builds them.


progenitor-impl/tests/output/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+pub mod buildomat_builder;
+pub mod buildomat_builder_tagged;
+pub mod buildomat_cli;
+//pub mod buildomat_httpmock;
```
jayvdb, line 4: commented out due to https://github.com/oxidecomputer/progenitor/issues/637

ahl:
What would you think about doing the `JsonSchema` -> `schemars::JsonSchema` and the file renames first?

jayvdb:
There is a lot of overlap, so would be good to get https://github.com/oxidecomputer/progenitor/pull/641 in quickly.

jayvdb:
This PR now sits on top of https://github.com/oxidecomputer/progenitor/pull/642

jayvdb:
@ahl , PR refreshed & green.

PR #642, opened 2023-12-08
Allow API to reside at submodule of crate

jayvdb:
Split off from https://github.com/oxidecomputer/progenitor/pull/638

Potentially `crate_name` args should be renamed.

### DIFF
--- a/progenitor-impl/src/cli.rs
+++ b/progenitor-impl/src/cli.rs
@@ -100,15 +100,20 @@ impl Generator {
             })
             .collect::<Vec<_>>();
 
-        let crate_ident = format_ident!("{}", crate_name);
+        let crate_path = syn::TypePath {
+            qself: None,
+            path: syn::parse_str(&crate_name).unwrap(),
+        };
 
         let code = quote! {
+            use #crate_path::*;
+
             pub struct Cli<T: CliOverride = ()> {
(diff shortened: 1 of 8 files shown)


ahl:
Is this fixing a bug or is this adding some new functionality? I'm not sure I understand the utility.

jayvdb:
This is new functionality, needed in order to have the generated cli & httpmock be a module that resides *beside* the generated client.

Currently the generator uses `format_ident!(..)` which only supports `foo` - this PR allows it to support `foo::bar`.

It is needed for https://github.com/oxidecomputer/progenitor/pull/638 , where the types+client, mocks and cli are three separate generated modules, and lib.rs is a manually authored file that combines those generated files and potentially adds other human authored functionality.  The top level lib.rs does not contain the types+client.  Therefore the mocks and cli need to be told where to find the types+client within the crate.

@franklin-ai is using this approach to be able to greater control over our crate lib.rs, re-exporting at the top level only a subset of the generated mods, and including lots of other human written code, such as impls on the generated types.

jayvdb:
ping @ahl . there are a bunch of other improvements which depend on this, some with a PR, others waiting on it getting merged.

ahl:
good enough; thanks

PR #643, opened 2023-12-08
Update buildomat sample OpenAPI

jayvdb:
Imported https://github.com/oxidecomputer/buildomat/blob/716350728338cf44f39a132213fd8f983283fb63/client/openapi.json


ahl:
Can we close this?

jayvdb:
I was wondering if it would be OK for me to split this file into a custom test case openapi file which include the scenarios which were added to buildomat , and then the file buildomat can be kept in sync with upstream ?

PR #655, opened 2023-12-13
Include Response bytes when payload didn't deserialize as expected

epompeii:
The current error handling for `progenatior-client` does not allow for reattempts at deserializing.
This is very useful when creating clients that will need a little more flexibility such as forwards compatibility.

This changeset breaks out the `InvalidResponsePayload` within `progenatior-client` into two separate errors:

- `InvalidResponseBytes`: When `reqwest` fails to actually get the bytes for the response.
- `InvalidResponsePayload`: When the payload didn't deserialize as expected, now also includes the bytes.

### DIFF
--- a/docs/progenitor-client.md
+++ b/docs/progenitor-client.md
@@ -74,7 +74,7 @@ There are five sub-categories of error covered by the error type variants:
 - An expected error response, defined by the OpenAPI document with a 4xx or 5xx
   status code
 
-- An expected status code (whose payload didn't deserialize as expected (this
+- An expected status code whose payload didn't deserialize as expected (this
   could be viewed as a sub-type of communication error), but it is separately
   identified as there's more information; note that this covers both success and
   error status codes
@@ -88,7 +88,8 @@ pub enum Error<E = ()> {
     InvalidRequest(String),
     CommunicationError(reqwest::Error),
     ErrorResponse(ResponseValue<E>),
-    InvalidResponsePayload(reqwest::Error),
+    InvalidResponseBytes(reqwest::Error),
+    InvalidResponsePayload(bytes::Bytes, reqwest::Error),
     UnexpectedResponse(reqwest::Response),
 }
 ```
--- a/docs/progenitor-client.md
+++ b/docs/progenitor-client.md
@@ -63,21 +63,24 @@ It can be used as the type `T` in most instances and extracted as a `T` using
 
 ## `Error<E>`
 
-There are five sub-categories of error covered by the error type variants:
+There are seven sub-categories of error covered by the error type variants:
 
-- A request that did not conform to API requirements. This can occur when
-  required builder or body parameters were not specified, and the error message
-  will denote the specific failure.
+- A request that did not conform to API requirements.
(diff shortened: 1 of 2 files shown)


progenitor-client/src/progenitor_client.rs
```diff
@@ -240,9 +240,14 @@ pub enum Error<E = ()> {
     /// A documented, expected error response.
     ErrorResponse(ResponseValue<E>),
 
+    /// An expected response when upgrading connection.
+    InvalidUpgrade(reqwest::Error),
+
+    /// An expected response code whose deserialization failed.
+    InvalidResponseBytes(reqwest::Error),
```
ahl, line 247: How about:
```suggestion
    /// Encountered an error reading the body for an expected response.
    ResponseBodyError(reqwest::Error),
```

progenitor-client/src/progenitor_client.rs
```diff
@@ -314,7 +323,13 @@ where
...
                 rve.fmt_info(f)
             }
-            Error::InvalidResponsePayload(e) => {
+            Error::InvalidUpgrade(e) => {
+                write!(f, "Invalid Response Upgrade: {}", e)
+            }
+            Error::InvalidResponseBytes(e) => {
+                write!(f, "Invalid Response Body Bytes: {}", e)
+            }
+            Error::InvalidResponsePayload(_b, e) => {
```
ahl, line 332: we don't want to emit the bytes?

progenitor-client/src/progenitor_client.rs
```diff
@@ -240,9 +240,14 @@ pub enum Error<E = ()> {
     /// A documented, expected error response.
     ErrorResponse(ResponseValue<E>),
 
+    /// An expected response when upgrading connection.
+    InvalidUpgrade(reqwest::Error),
```
ahl, line 244: Let's move this up to be right after `CommunicationError` which is probably what this should have been initially.

```suggestion
    /// An error occurred when upgrading connection.
    UpgradeError(reqwest::Error),
```

docs/progenitor-client.md
```diff
@@ -88,7 +88,8 @@ pub enum Error<E = ()> {
     InvalidRequest(String),
     CommunicationError(reqwest::Error),
     ErrorResponse(ResponseValue<E>),
```
ahl, line 90: let's add the upgrade error here as well

progenitor-client/src/progenitor_client.rs
```diff
@@ -6,7 +6,7 @@
 
 use std::ops::{Deref, DerefMut};
 
-use bytes::Bytes;
+pub use bytes::Bytes;
```
ahl, line 9: does this need to be pub? `Error::InvalidResponsePayload`'s inner type isn't `pub`

ahl (changes requested):
This is great; thanks for doing this!

epompeii:
Thanks for all the feedback, @ahl !
I have implemented all the suggested changes. Let me know if you see anything else 😃

ahl (approved):

ahl:
Thanks for this!

Issue #664, opened 2023-12-28
Feature request: granular code generation


daniel-levin:
Hello there! I am a happy user of this crate. I have a feature request, which I want to gauge interest for. I am happy to implement it myself, but I want to solicit some feedback before submitting a PR.

Today, progenitor generates all the code you need to build reqwest-based clients. I have a use case that does not require the actual reqwest client. Instead, it requires only the generated structs and enums. Namely, the code inside the `types` module.

The use case is: I want to use OpenAPI to define objects but I don't necessarily want to use them in HTTP APIs.

There are several legitimate uses for this:

1. Putting objects onto a message queue.
2. Putting objects into persistent storage / utility blob storage.
3. Tokio is unavailable

First, I'll explain what I don't want to do. It would be trivial to add a field to `GenerationSettings` which would instruct progenitor to only generate types. This would solve my immediate problem but it won't be clear to those that come after why `types` gets special treatment.

Instead, I propose that `GenerationSettings` is re-jigged so that it can generate only the things you want. By default, it will continue to generate everything, and the API won't change for existing users. But, you can opt-in/opt-out of certain features.

It would be great if you could, say, do this:
```rust
GenerationSettings::bare()
.with_types();
```

and it would generate only the `types` module.

This technique would also allow progenitor to add orthogonal features without disrupting existing uses. For example, you might want something like this:

```rust
GenerationSettings::bare()
.with_types()
.with_blocking_client()
.with_comprehensive_logging(RegulationEnvironment::Gdpr);
```

and so on.

daniel-levin:
I've been prototyping what this would look like here: https://github.com/daniel-levin/progenitor/tree/granular-generation

I haven't implemented granular code generation yet, but it would be very natural to swap out a reqwest client for an alternative one.

ahl:
I'm going to need to spend more time on this, but in its face: I like the concept a lot. Agreed on the utility of emitting just types; agreed on the potential need for different clients. Conversely, adding configuration options is a slippery slope. I don't want this to turn into openapi-generator which is so sprawling and ambitious that it is effectively untestable. In other words, there's a balance between "configurable" and "opinionated" that we want to consider carefully--in particular in the face of work we know we want to take on in the future.

Thanks. I'll try to get to this in the next week or two, but feel free to ping me (here, discord, mastodon, twitter) if you're growing impatient.

[...]

PR #666, opened 2023-12-29
Add flag to generate only the types module.

daniel-levin:
See #664 

This feature adds a new method to `GenerationSettings` which tells progenitor to only generate the types module. I've added an example crate showing how to use this, and why it might be useful.


daniel-levin:
Closing in favor of #668

PR #668, opened 2023-12-31
Add a layer of indirection to the generation of the types module, which can also be used independently.

daniel-levin:
As it stands, it is not possible to generate only types specified in the OpenAPI document. That is, you can't only generate the `types` module, without the associated client.  See: #664. This change does two things:

1. It introduces a new public API to progenitor that makes it easy to generate only types.
2. It slightly simplifies the code generator, which will support the eventual project to generate clients that don't use reqwest, while not changing the code generator's API at all.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -81,7 +81,7 @@ jobs:
         default: true
         target: wasm32-unknown-unknown
     - name: Install wasm-bindgen-cli
-      run: cargo install --version 0.2.87 wasm-bindgen-cli
+      run: cargo install --version 0.2.89 wasm-bindgen-cli
     - name: Run tests
       run: |
         cd example-wasm
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.41",
+ "syn 2.0.43",
 ]
 
 [[package]]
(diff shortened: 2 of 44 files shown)


daniel-levin:
@ahl When you get a moment, please can you take a look?

ahl:
@daniel-levin I hope to take a look this weekend; thanks for your patience.

ahl:
Looking at this more closely and at the previous PR, I need to be candid: while I liked the concept, I think in practice it adds complexity and burden in areas that will likely undergo change. If there's a way to satisfy your use case with the existing interfaces I'd like to help you find it.

For example, could you do something like this:

```rust
    let file: syn::File =
        syn::parse2(tokens).expect("progenitor handed us bad code!");
    let types = file
        .items
        .iter()
        .find(|item| matches!(item, syn::Item::Mod(mod_item) if mod_item.ident == "types"))
        .expect("there was no types mod");
    let new_tokens = types.to_token_stream();
```

I'm already unhappy with the interface to Generator: `cli`, `httpmock`, `generate_tokens`. I think that `TypeGenerator::generate_types_module_tokens` further complicates the library. If something like what I suggest doesn't work for you, let me think a little harder about how to make this more palatable?

daniel-levin:
Ahhhh, I finally understand what you mean. Yeah, that'll solve my problem.

> I'm already unhappy with the interface to Generator

I wanted to make my change fully backwards compatible to _preserve_ this interface. But, given that you don't like it, it's subject to change, and my immediate problem is solved, I'm going to cancel this PR. Thanks for the prompt and useful feedback.

ahl:
@daniel-levin thank **you** for your contribution and patience. I'm trying not to let an uncertain future and my rather abstract aesthetic concerns limit your ability to use this repo.

PR #708, opened 2024-02-03
CLI: add an output interface for consumers

ahl:
See its use here: https://github.com/oxidecomputer/oxide.rs/pull/545


ahl:
> This as opposed to implementing `CliConfig` for each output format.

My expectation is that different output formats would be implemented by the consumer with a single impl of CliConfig per consumer. That is to say, I'd expect the CliConfig to pass along the format to a particular formatter.

I'd like to have a single interface between generated code and the consumer of it.

PR #724, opened 2024-02-16
Allow query parameters to be an array

MichielHegemans:
Does not support `explode: false` at the moment as the `.query(&query)` call does urlencode on the `,` in case we join the arguments.

It is not entirely clear to me whether or not it is correct, but I believe that `id=a%2Cb` results into `id = ["a,b"]` while `id=a,b` results in `id = ["a", "b"]`

It does support `type: array` in the openapi spec and will add an exploded version to the URL `id=a&id=b` to the query.


ahl:
Obviated by #1017

PR #729, opened 2024-02-21
update generated code for Rust 1.76

sunshowers:
In some cases (e.g. ddm-admin-client and mg-admin-client in omicron), the
progenitor-generated file was not used at the top level. With Rust 1.76, `pub
use` for a type that's not actually exported publicly warns as well.


ahl (approved):
ship it!

PR #743, opened 2024-03-16
Use toml instead of diff for toml code snippets

matthiasbeyer:
Use the toml syntax highlighter and format the code snippets as toml instead of diff, so users can copy-paste it more easily, without having to edit out the diff makers (+) by hand when they add the snippets to their Cargo.toml / etc.


ahl (approved):
I'm sold

PR #756, opened 2024-04-01
Fix dropshot-pagination `_stream` client codegen

### LINKED ISSUE #755: Client paginated `_stream()` methods are confused about `limit`
@bnaecker and I ran into this while investigating https://github.com/oxidecomputer/omicron/issues/5364. Looking at this macro-expanded code from `progenitor::generate_api!`:

```rust
    /**List all metric producers assigned to an oximeter collector as a Stream

Sends repeated `GET` requests to `/metrics/collectors/{collector_id}/producers` until there are no more results.

Arguments:
- `collector_id`: The ID of the oximeter collector.
- `limit`: Maximum number of items returned by a single call
- `sort_by`
*/
    pub fn cpapi_assigned_producers_list_stream<'a>(
        &'a self,
        collector_id: &'a uuid::Uuid,
        limit: Option<std::num::NonZeroU32>,
        sort_by: Option<types::IdSortMode>,
    ) -> impl futures::Stream<
        Item = Result<types::ProducerEndpoint, Error<types::Error>>,
    > + Unpin + '_ {
        use futures::StreamExt;
        use futures::TryFutureExt;
        use futures::TryStreamExt;
        let final_stream_limit = limit
            .clone()
            .and_then(|x| std::num::NonZeroUsize::try_from(x).ok())
            .map(std::num::NonZeroUsize::get)
            .unwrap_or(usize::MAX);
        self.cpapi_assigned_producers_list(collector_id, limit, None, sort_by)
            .map_ok(move |page| {
                let page = page.into_inner();
                let first = futures::stream::iter(page.items).map(Ok);
                let rest = futures::stream::try_unfold(
                        page.next_page,
                        move |state| async move {
                            if state.is_none() {
                                Ok(None)
                            } else {
                                self.cpapi_assigned_producers_list(
                                        collector_id,
                                        None,
                                        state.as_deref(),
                                        None,
                                    )
                                    .map_ok(|page| {
                                        let page = page.into_inner();
                                        Some((
                                            futures::stream::iter(page.items).map(Ok),
                                            page.next_page,
                                        ))
                                    })
                                    .await
                            }
                        },
                    )
                    .try_flatten();
                first.chain(rest)
            })
            .try_flatten_stream()
            .take(final_stream_limit)
            .boxed()
    }
```

it looks like there is internal confusion about what `limit` is supposed to be:

1. It is converted into `final_stream_limit` and then applied to the returned stream via `.take(final_stream_limit)`
2. It is passed to the first call to `cpapi_assigned_producers_list` as the page limit, but...
3. ...it is _not_ passed to all subsequent calls to `cpapi_assigned_producers_list`

I think because of 1, this function does _not_ do what it claims; it will only fetch at most `limit` items. That seems wrong, and also `final_stream_limit` seems superfluous? Clients can call `.take(_)` on the returned stream themselves if they want to apply a final limit.

I'll start working on a test for this and a change that will (a) remove `final_stream_limit` and (b) pass `limit` to every invocation of the inner `_list()` call.

jgallagher:
Closes #755; see it for details!


progenitor-impl/src/method.rs
```diff
@@ -710,9 +710,12 @@ impl Generator {
             let step_params = method.params.iter().map(|param| {
                 if param.api_name.as_str() == "page_token" {
                     quote! { state.as_deref() }
-                } else if let OperationParameterKind::Query(_) = param.kind {
-                    // Query parameters are None; having page_token as Some(_)
-                    // is mutually exclusive with other query parameters.
+                } else if param.api_name.as_str() != "limit"
```
jgallagher, line 713: This is the change I'm least sure about, but I _think_ it's fine? We should only be here if we have a non-`None` `dropshot_pagination`, and `dropshot_pagination_data` explicitly checks for both `"page_token"` and `"limit"` as query parameters.

bnaecker (approved):

progenitor-impl/src/method.rs
ahl: Yeah, this is the part I'm least sure about as well. My intention (unexpressed in any comments) is that the consumer might specify the size of the first page to optimize for latency and then rely on the server to give us chunky, throughput oriented pages subsequently. Imagine, for example, a CLI where we  want to start printing output quickly.

Alternatively, would it make sense to have parameters first the first page and other pages?

In either case, if we don't include this change we might consider updating the function parameters docs which, currently, come from dropshot via the OpenAPI doc.

ahl (commented):
Progenitor also emits the CLI. My intention with the `--limit` parameter for various `list` subcommands is that the limit impacts the total number of items output--not just the particular page size.

I expect we'd want to honor the limit by generating a take there (i.e. rather than in the obviously wrong position in the generated stream function).

Would it be superfluous to update the docs for these generated _stream() / .stream() functions to point consumers to `.take()` as a way to impose an outer limit or should that be implicit in the use of the stream (not sure why that didn't occur to me but FWIW it didn't).

progenitor-impl/src/method.rs
jgallagher: > Alternatively, would it make sense to have parameters for the first page and other pages?

I thought about this. It doesn't seem unreasonable to me, although I suspect most clients _ought_ to specify `Some(small_value)` for the first page and `None (server choice)` for the other pages, right? I wonder if it would help to put this in a new type specifically for these parameters for clarity? Totally spitballing, something like:

```rust
pub struct StreamPaginationLimits {
    // or just omit this one and recommend clients use `.take(_)`
    pub total: Option<NonZeroU32>,
    pub first_page: Option<NonZeroU32>,
    pub remaining_pages: Option<NonZeroU32>,
}
```

with helper methods for "let the server pick all the limits, I don't care" and "I just want a small first page, let the server pick for remaining pages".

jgallagher: Another thought: I'm not sure the stream interface actually does the "small first page, larger subsequent pages" thing you want, although I guess it depends on the client. I _think_ if you're using

```rust
while let Some(item) = stream.next() { .. }
```

with a first page limit of 10 and a subsequent page limit of 100, you'll incur the latency for the `GET (10)` on the first `nex()`, the next 9 will return immediately, then you'll incur the latency for `GET (100)` on the 11th call to `next()`, right?

ahl (commented):
there's similar code in method::builder_struct -- can you please update that similarly?

jgallagher:
> there's similar code in method::builder_struct -- can you please update that similarly?

Done in [faf3f72](https://github.com/oxidecomputer/progenitor/pull/756/commits/faf3f7239e8f114067c327d792c3e57975fb7cd8)

ahl:
Looks good. Thanks

PR #757, opened 2024-04-02
Revert reqwest update from 0.11 -> 0.12

bnaecker:
This causes lots of downstream upgrade challenges, because the Progenitor version has not changed but `reqwest` versions are incompatible.


ahl (approved):
"Marty, you've got to come back with me!"

PR #782, opened 2024-04-20
Mock code doesn't compile if path parameters are renamed

### LINKED ISSUE #780: Mock client can't handle path parameters with capitals
I pretty much copied a schema from https://swagger.io/docs/specification/paths-and-operations/, but I changed `id` in the path parameter to `iD`. 

<details>

<summary>JSON</summary>

```
{
  "openapi": "3.0.1",
  "info": {
    "title": "title",
    "version": "version"
  },
  "paths": {
    "/users/{iD}": {
      "get": {
        "tags": [
          "Users"
        ],
        "summary": "Gets a user by ID.",
        "description": "A detailed description of the operation. Use markdown for rich text representation, such as **bold**, *italic*, and [links](https://swagger.io).\n",
        "operationId": "getUserById",
        "parameters": [
          {
            "name": "iD",
            "in": "path",
            "description": "User ID",
            "required": true,
            "schema": {
              "type": "integer",
              "format": "int64"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "Successful operation",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/User"
                }
              }
            }
          }
        },
        "externalDocs": {
          "description": "Learn more about user operations provided by this API.",
          "url": "http://api.example.com/docs/user-operations/"
        }
      }
    }
  },
  "components": {
    "schemas": {
      "User": {
        "type": "object",
        "properties": {
          "id": {
            "type": "integer",
            "format": "int64"
          },
          "name": {
            "type": "string"
          }
        },
        "required": [
          "id",
          "name"
        ]
      }
    }
  }
}
```

</details>

Then I'm getting this error compiling the mock client:
```
error: argument never used
  --> catalog-api-v1/src/mock.rs:20:64
   |
20 |             let re = regex::Regex::new(&format!("^/users/.*$", value.to_string()))
   |                                                 -------------  ^^^^^^^^^^^^^^^^^ argument never used
   |                                                 |
   |                                                 formatting specifier missing
```

Afaict, it's assumed that `as_wildcard_param` always returns a `String` with `{}` in it, but currently that's not the case.

ahl:
This occurs if path parameters are renamed e.g. because they aren't `lower_snake_case` or if they conflict with Rust keywords (e.g. `type`).

Fixes #780

@mkenigs I think this addresses the problem you raised; please let me know if this doesn't address it.

### DIFF
--- a/progenitor-impl/src/httpmock.rs
+++ b/progenitor-impl/src/httpmock.rs
@@ -191,7 +191,7 @@ impl Generator {
                 let name_ident = format_ident!("{}", name);
                 let handler = match kind {
                     OperationParameterKind::Path => {
-                        let re_fmt = method.path.as_wildcard_param(name);
+                        let re_fmt = method.path.as_wildcard_param(api_name);
                         quote! {
                             let re = regex::Regex::new(
                                 &format!(#re_fmt, value.to_string())
--- a/progenitor-impl/src/httpmock.rs
+++ b/progenitor-impl/src/httpmock.rs
@@ -161,7 +161,11 @@ impl Generator {
         // can specify a prescribed value for that parameter.
         let when_methods = method.params.iter().map(
             |OperationParameter {
-                 name, typ, kind, ..
+                 name,
+                 typ,
+                 kind,
+                 api_name,
+                 description: _,
              }| {
                 let arg_type_name = match typ {
                     OperationParameterType::Type(arg_type_id) => self
--- a/progenitor-impl/tests/output/src/buildomat_httpmock.rs
+++ b/progenitor-impl/tests/output/src/buildomat_httpmock.rs
@@ -84,7 +84,7 @@ pub mod operations {
         }
 
         pub fn task(self, value: &str) -> Self {
-            let re = regex::Regex::new(&format!("^/v1/task/{}$", value.to_string())).unwrap();
+            let re = regex::Regex::new(&format!("^/v1/task/.*$", value.to_string())).unwrap();
             Self(self.0.path_matches(re))
         }
     }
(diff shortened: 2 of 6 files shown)


progenitor-impl/tests/output/src/buildomat_httpmock.rs
```diff
@@ -84,7 +84,7 @@ pub mod operations {
         }
 
         pub fn task(self, value: &str) -> Self {
-            let re = regex::Regex::new(&format!("^/v1/task/{}$", value.to_string())).unwrap();
+            let re = regex::Regex::new(&format!("^/v1/task/.*$", value.to_string())).unwrap();
```
ahl: this won't compile

progenitor-impl/tests/output/src/buildomat_httpmock.rs
```diff
@@ -84,7 +84,7 @@ pub mod operations {
         }
 
         pub fn task(self, value: &str) -> Self {
-            let re = regex::Regex::new(&format!("^/v1/task/.*$", value.to_string())).unwrap();
+            let re = regex::Regex::new(&format!("^/v1/task/{}$", value.to_string())).unwrap();
```
ahl, line 87: ... this does compile.

progenitor-impl/src/httpmock.rs
```diff
@@ -191,7 +191,7 @@ impl Generator {
                 let name_ident = format_ident!("{}", name);
                 let handler = match kind {
                     OperationParameterKind::Path => {
-                        let re_fmt = method.path.as_wildcard_param(name);
+                        let re_fmt = method.path.as_wildcard_param(api_name);
```
ahl, line 194: we want to lookup the original name not the name of the generated identifier

PR #823, opened 2024-06-06
add mut request `pre_hook`

drahnr:
Currently `progenitor` supports `pre_hook`, `pre_hook_async` and `post_hook`, where the further get a `&reqwest::Request` passed.
The auth mechanism for an `OpenAPI` spec requires to _sign_ the encoded query and append two key value pairs to the request, as such the `pre-hook` appears to be the right place to adjust the `Request` accordingly.

Proposal:

Add `with_pre_hook_mut` in addition, allowing to modify the `Request` accordingly.

Considerations:

Since it's a per-request signature, modifying the `Client` itself is not sufficient.

Altneratives:

Change the API of the current `pre_hook` to take a `&mut` and use lint exception to avoid getting warnings on existing code.

### DIFF
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -1134,6 +1134,11 @@ impl Generator {
                 (#hook)(&#client.inner, &#request_ident);
             }
         });
+        let pre_hook_mut = self.settings.pre_hook_mut.as_ref().map(|hook| {
+            quote! {
+                (#hook)(&#client.inner, &mut #request_ident);
+            }
+        });
         let pre_hook_async = self.settings.pre_hook_async.as_ref().map(|hook| {
             quote! {
                 match (#hook)(&#client.inner, &mut #request_ident).await {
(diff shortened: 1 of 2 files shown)


jclulow:
FWIW, when I added `pre_hook_async` it was also for signing requests in a similar manner, and as such already gets a mutable reference to the request.

drahnr:
> FWIW, when I added `pre_hook_async` it was also for signing requests in a similar manner, and as such already gets a mutable reference to the request.

I'd be in favor to closing this PR and changing the sync hook API to take a `&mut Request`.

progenitor-impl/src/method.rs
```diff
@@ -1134,6 +1134,11 @@ impl Generator {
                 (#hook)(&#client.inner, &#request_ident);
```
ahl, line 1134: let's just make this mut; it was probably a mistake to diverge the non-async and async variants. Let's not compound the mistake.

ahl (changes requested):

PR #825, opened 2024-06-08
fix: pre_hook (sync) make request mutable

drahnr:
Obsoletes #823

### DIFF
--- a/progenitor-impl/src/method.rs
+++ b/progenitor-impl/src/method.rs
@@ -1131,7 +1131,7 @@ impl Generator {
 
         let pre_hook = self.settings.pre_hook.as_ref().map(|hook| {
             quote! {
-                (#hook)(&#client.inner, &#request_ident);
+                (#hook)(&#client.inner, &mut #request_ident);
             }
         });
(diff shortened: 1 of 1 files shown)


ahl (commented):
I'm sort of surprised that we don't have a test that validates this. What will existing users of `pre_hook` experience other than a type error if they've specified the type in their pre_hook closure or fn?

drahnr:
> I'm sort of surprised that we don't have a test that validates this. What will existing users of `pre_hook` experience other than a type error if they've specified the type in their pre_hook closure or fn?

Can you rephrase, I am not sure I understand what your concern is. Do you want an interface check for the provided `TokenStream` or are you concern re change in behviour?

ahl:
If there is an existing user of the `pre_hook` interface, what will those users experience with this change? Will everything just work or will compilation fail with an error?

drahnr:
Happy new year!

Mildly belated response; the user won't see a delta, since a we can call any `fn foo(&T)` with `foo(&mut t)`.

ahl (approved):
thanks for this; sorry for the very long delay

PR #833, opened 2024-06-19
chore: bump reqwest to 0.12.5

gusinacio:


ahl:
We've had dependabot ignoring this because we aren't yet ready to make the leap to hyper v1 in our various repositories. The first step is https://github.com/oxidecomputer/dropshot/pull/1028 which is hopefully close. Until that time, we're going to want to stick with reqwest 0.11.

citizen-stig:
Looking forward to that!

PR #856, opened 2024-07-09
add an `api-version` header to all requests

### LINKED ISSUE #564: progenitor clients should specify API version

ahl:
Closes #564.

### DIFF
--- a/progenitor-impl/src/lib.rs
+++ b/progenitor-impl/src/lib.rs
@@ -452,6 +452,14 @@ impl Generator {
                         let dur = std::time::Duration::from_secs(15);
 
                         reqwest::ClientBuilder::new()
+                            .default_headers(
+                                [(
+                                    reqwest::header::HeaderName::from_static("api-version"),
+                                    reqwest::header::HeaderValue::from_static(#version_str),
+                                )]
+                                .into_iter()
+                                .collect(),
+                            )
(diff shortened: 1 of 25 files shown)


progenitor-impl/src/lib.rs
```diff
@@ -452,6 +452,14 @@ impl Generator {
                         let dur = std::time::Duration::from_secs(15);
 
                         reqwest::ClientBuilder::new()
+                            .default_headers(
```
davepacheco, line 455: I think this won't cover the `new_with_client()` case, right?

ahl: That does seem probably bad. My thought here--as I recall--was that it felt rude to modify a `reqwest::Client` that the user had provided to us, but I suppose that's both overly fussy **and** missing the most relevant cases. Thanks for taking a look.

ahl:
@davepacheco I think I need to change this progenitor interface more significantly--once we have a `reqwest::Client` e.g. from `new_with_client` there isn't anything we can do to glue in new our header. Some options:

```rust
// 1
Client::new_with_builder(reqwest::Client::builder().custom_settings());
// 2
Client::new_map(|builder| builder.customsettings());
```

Are there others? I think I prefer 2 in that it (I think) makes it clear to the user that the `Client` is in charge and is allowing you to make changes.

PR #864, opened 2024-07-18
Make Clap sort subcommands automatically

wfchandler:
By default `clap` will use the order subcommands are added to a `Command` to determine print order in `--help`. In our case this results in the help text being randomly sorted, which is a confusing experience for users.

Add `.display_order(0)` when creating a new `Command` so that `clap` will automatically sort the subcommands for us.

### DIFF
--- a/progenitor-impl/src/cli.rs
+++ b/progenitor-impl/src/cli.rs
@@ -214,6 +214,8 @@ impl Generator {
             }
         });
 
+        let display_order = quote! { .display_order(0) };
+
         let fn_name = format_ident!("cli_{}", &method.operation_id);
 
         let cli_fn = quote! {
@@ -223,6 +225,7 @@ impl Generator {
                 #parser_args
                 #about
                 #long_about
+                #display_order
             }
         };
 
--- a/progenitor-impl/tests/output/src/buildomat_cli.rs
+++ b/progenitor-impl/tests/output/src/buildomat_cli.rs
@@ -34,24 +34,26 @@ impl<T: CliConfig> Cli<T> {
     }
 
     pub fn cli_control_hold() -> clap::Command {
-        clap::Command::new("")
+        clap::Command::new("").display_order(0)
     }
 
     pub fn cli_control_resume() -> clap::Command {
-        clap::Command::new("")
+        clap::Command::new("").display_order(0)
     }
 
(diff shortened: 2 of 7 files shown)


wfchandler:
For the purposes of `oxide.rs` this PR isn't necessary, the change in https://github.com/oxidecomputer/oxide.rs/pull/751 is enough to sort all subcommands. However, I think this is still a useful default for anyone else using `progenitor`.

ahl:
Would it be possible for consumers to choose the ordering themselves without generating this code?

wfchandler:
> Would it be possible for consumers to choose the ordering themselves without generating this code?

@ahl yes, the consumer could manually iterate over the generated commands and update the display order as desired, this is what https://github.com/oxidecomputer/oxide.rs/pull/751 did.

ahl:
Then I suggest we leave this to consumers.

wfchandler:
> Then I suggest we leave this to consumers.

Why stick with a bad default given the low level of effort needed on our part?

This change provides a reasonable default that the large majority of consumers of this library will want. In the unusual case that they do want to define a custom order, this change does nothing to prevent them from doing so.

wfchandler:
> > > Then I suggest we leave this to consumers.
> > 
> > 
> > Why stick with a bad default given the low level of effort needed on our part?
> 
> Why is the current behavior bad?

Programmatically generated commands are unlikely to be created in lexical order. Claps lists the `--help` subcommands in the order they were added unless `display_order` is explicitly set. As a result, default help output will most likely be in random order, and harder to follow.

ahl:
They don't appear to be in a random order. Each command is processed in the order that the corresponding operations appear in the input OpenAPI document. More specifically, the `CliCommand::iter` method produces an iterator in that order and then the consumer (e.g. oxide.rs) slots those into subcommands. Is this order meaningful? Maybe? I don't know. Given than consumers need to iterate over commands (at least currently), I'm inclined to leave this to the consumer.

Do you think there's a strong rationale for doing it in the generated code? I confess that preserving ordering within the OpenAPI document isn't a strong rationale, but it is an ordering that's otherwise unrecoverable.

wfchandler:
I suspect the number of OpenAPI docs ordered with CLI help is mind is not high, but can't say it's an impossibility. Will close this out.

PR #871, opened 2024-07-28
Bump `time` to latest for rust 1.80.0 compatibility.

pvandommelen:
The project no longer compiles on 1.80.0 because of a missed type inference in the `time` library.


ahl:
I was able to build with 1.80.0. Can you share the failure you're seeing and how you're encountering it?

ahl:
Never mind; I see it now.

PR #872, opened 2024-07-28
Basic support for vector query parameters.

### LINKED ISSUE #692: Query param of `type: array` with `items.type: string` leads to `to_string()` and unsatisfied trait on `&&Vec<String>`
While working with Scaleways API spec for [Accounts](https://www.scaleway.com/en/developers/api/account/#path-projects-list-all-projects-of-an-organization), I encountered a array of strings in the URL. To me it is not clear how exactly it should work. The doc suggests it ends up as string in the URL `&project_ids=[string]`, so I assume they want it joined with a `,` and not `&project_ids=1&project_ids=2`. 

From Swaggers [docs on OpenAPI serialization](https://swagger.io/docs/specification/serialization/), the default style is `style: form` and `explode: true` aka `&project_ids=1&project_ids=2`. That would, I think, require a change here https://github.com/oxidecomputer/progenitor/blob/91564cf74fbc24c5fb1af0439c8371dc8bbea304/progenitor-impl/src/method.rs#L843 to account for `Vec<_>` and iterate + push for that case.

Any thoughts?

pvandommelen:
Includes a small refactor which changes generated variable names in the output.

This currently assumes explode=true, and therefore only partially resolves #692. I'm looking for some guidance on how to best improve this. The openapiv3 crate exposes `explode` in its ParameterData so the necessary configuration is available, but it's unclear to me if this value is best passed separately as part of the current `OperationParameter`, or within `OperationParameterKind::Query`.

While I have manually tested this with an endpoint another issue is the lack of a test. The `oxidecomputer/dropshot` server used in  tests doesn't yet support non-scalar query parameters. Another option is to include an openapi document and verify its output, but that feels rather heavy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1203,6 +1203,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "num-conv"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+
+[[package]]
 name = "num-traits"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


ahl:
Obviated by #1017. Thank you for this contribution!

PR #897, opened 2024-08-23
Reduce dependency tree a little bit

citizen-stig:
Don't import default features for `clap` and `chrono`.

Progenitor is a great crate and used by my team and we aiming for faster compilation, so reducing features imported by default helps that.


ahl:
Thanks for this; good idea.

PR #929, opened 2024-09-29
fix: specify serde_json required for build.rs dependencies

b5:
Encountered this during setup, if it doesn't generalize please feel free to ignore


ahl:
This is great. Thanks for the contribution.

PR #941, opened 2024-10-07
Improve handling of name collisions between parameters and top-level body fields

ahl:
In https://github.com/oxidecomputer/oxide.rs/pull/853 we had a situation where a parameter (query) had the same name as a top-level field in the POST body. In this case, we prioritized the body field over the parameter. This led to the CLI subcommand being unusable. While this particular case turned out to be a mistake in the API, if such a situation were to occur, this PR causes us to ignore the body field (thus making the body required if the field was required).

This PR has a commit that introduces the new test for this condition and then a commit that contains the fix.

### DIFF
--- a/progenitor-impl/src/cli.rs
+++ b/progenitor-impl/src/cli.rs
@@ -614,13 +614,13 @@ impl Generator {
 
         let scalar = prop_type.has_impl(TypeSpaceImpl::FromStr);
 
-        if scalar {
+        let prop_name = name.to_kebab_case();
+        if scalar && !args.has_arg(&prop_name) {
             let volitionality = if required {
                 Volitionality::RequiredIfNoBody
             } else {
                 Volitionality::Optional
             };
-            let prop_name = name.to_kebab_case();
             let parser = clap_arg(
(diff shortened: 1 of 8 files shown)


rcgoodfellow:
> this PR causes us to ignore the body field (thus making the body required if the field was required)

I'm trying to internalize this statement. I would have thought the body would be required if the _parameter_ was required e.g.:

> this PR causes us to ignore the body field (thus making the body required if the parameter was required)

?

ahl:
Let's say the body has a top-level, scalar field whose name conflicts with that of a parameter (query or path). In that situation we would previously just fill in that field of the body and not supply the parameter which was a big problem! (particularly if the parameter was required)

After this change, in such a case, the generated code ONLY fills in the parameter. If the field is required (ie not an Option type) that will make the --json-body parameter mandatory as it would be the only way to supply that required field. This is similarly the case if a required body field is not a scalar and therefore cannot be expressed as a simple parameter on the command line.

rcgoodfellow (approved):
Thanks for the clarification. Changes LGTM!

PR #942, opened 2024-10-09
add async post_hook function support

upachler:
Currently, the builder offers `with_pre_hook()`, `with_pre_hook_async()` and `with_post_hook()`. However, there is no async variant for post call hooks. This PR adds that hook function.

I actually found I needed it because I needed to act on HTTP responses by sending a message via a tokio mpsc channel - and that is only possible within an async function context.


ahl (approved):

ahl:
Thanks for this.

PR #943, opened 2024-10-09
Return source of wrapped errors

wfchandler:
The `Display` impl for `Error` wraps the original error received with a message, e.g., "Communication error: {original_error_message}". The `source` method returns the same error, though, so iterating over the error chain will cause the same error message to be printed twice.

Update `source` to return the source error of the original error, not our wrapper, to avoid this duplication.


ahl:
I'm not sure I agree with this. Based on what we learned wouldn't it be more accurate to not print information about the source in the Display impl?

wfchandler:
Agreed, this is not the right approach.

PR #949, opened 2024-10-14
Show source chain with alt format for Error

wfchandler:
For debugging purposes it can be useful to expose the full details of an error, but this output is often too verbose to show in the default Display format. The anyhow crate is a widely used example of this approach, exposing the full backtrace chain when the `{:#}` alternate formatting modifier is used.

Add an alternate Display formatting modifier for Error that will show the full source chain. We skip the first source, as this is already printed as part of the outermost error.


wfchandler:
I intend to use this for debug logging in `oxide.rs`.

PR #960, opened 2024-10-28
E does not need to implement std::fmt::Debug in ResponseValue::from_response

hepek:
I was getting compilation errors in my progenitor generated code when the generated error type is `Error<ByteStream>`, because ByteStream does not implement std::fmt::Debug.

I cannot find any evidence that E needs to impl std::fmt::Debug from this function.

So I removed the trait bound from it.


ahl:
Thanks for this!

PR #965, opened 2024-11-04
Subschemas cause the value of `nullable` to be ignored

ahl:
Let's say we have a schema like this:

```json
{
  "type": "object",
  "required": ["things"],
  "properties": {
    "things": {
      "nullable": true,
      "allOf": [
        {
          "$ref": "#/components/schemas/Foo",
        }
      ]
    }
  }
}
```

We'd expect to get this type:

```rust
struct SomeType {
    things: Vec<Option<Foo>>,
}
```

But we don't! Instead we get:

```rust
struct SomeType {
    things: Vec<Foo>,
}
```

Where did the `Option` go? It turns out this isn't a `typify` bug--rather there's a bug in progenitor related to the conversion of OpenAPI schemas to the JSON schema representation. In particular, when dealing with types that 1. have no `type` field 2. contain subschemas (`oneOf`, `allOf`, `anyOf`, `not`) and 3. contain `nullable: true`. In these situations, `nullable` is effectively ignored.

This change improves the handling of that situation. In particular, if `nullable` is `true` and--let's say--the output schema was be `X`, now we do (more or less):

```json
{
  "oneOf": [
    { "type": "null" },
    X
  ]
}
```

Once we augment the conversion in this way, typify produces the type we expect.

### DIFF
--- a/progenitor-impl/src/to_schema.rs
+++ b/progenitor-impl/src/to_schema.rs
@@ -312,7 +312,7 @@ impl Convert<schemars::schema::Schema> for openapiv3::Schema {
                 ..Default::default()
             },
 
-            openapiv3::SchemaKind::OneOf { one_of } => {
+            openapiv3::SchemaKind::OneOf { one_of } => oneof_nullable_wrapper(
                 schemars::schema::SchemaObject {
                     metadata,
                     subschemas: Some(Box::new(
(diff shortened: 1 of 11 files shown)


augustuswm (approved):

ahl:
Note that we're already suffering (silently) from this bug in the Oxide SDK generation, but I noticed the problem here: https://github.com/oxidecomputer/oxide.rs/pull/899/commits/3d68438b35c7838f600c834353f5ae59989597b2

PR #991, opened 2024-12-03
refactor(error): wrapped error responses print using debug formatting

JosiahBull:
## Reasoning

Debug formatting is preferred over Display as it will also print information about wrapped errors, including stack traces, which can make debugging much easier.

Note that this was already the case with `UnexpectedResponse` variant, this change standardises it across all error variants.

## Alternatives

`Debug` formatting in a `Display` implementation is a bit of an anti-pattern. The `Error` type is using the same string formatting implementation for both it's `Display` and `Debug` traits. I can split up this functionality so that the `Display` information can be slightly less verbose if this is desired.

### DIFF
--- a/progenitor-client/src/progenitor_client.rs
+++ b/progenitor-client/src/progenitor_client.rs
@@ -330,17 +330,17 @@ where
                 write!(f, "Invalid Request: {}", s)?;
             }
             Error::CommunicationError(e) => {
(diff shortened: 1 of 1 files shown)


ahl:
I don't really understand the motivation. In what situation would this be helpful?

ahl:
Feel free to re-open if you'd like to proceed with this.

PR #1011, opened 2024-12-23
feat(map_type): support custom map types.

### LINKED ISSUE #985: Support Generating IndexMap Instead of HashMap
For some usecases, e.g. snapshots in tests, consistent ordering via `IndexMap` would be preferred over a randomly ordered `HashMap` in generated clients.

I am happy to do the work required to allow users of the `progenitor` library to opt-in to this feature via a feature flag or as the default. I can alternatively extend support this feature via the `x-rust-type` mechanism, with a special case when `path` is equal to `indexmap::IndexMap` or `indexmap::IndexSet` to enable this feature.

Let me know if this is in-scope for the features you'd like `progenitor` to support and I'll open a PR. 👍

JosiahBull:
Closes #985

@ahl Let me know what more you'd like to see here, kept the initial impl pretty bare bones as last time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2332,7 +2332,7 @@ checksum = "dcf81ac59edc17cc8697ff311e8f5ef2d99fcbd9817b34cec66f90b6c3dfd987"
 [[package]]
 name = "typify"
 version = "0.2.0"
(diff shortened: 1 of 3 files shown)


ahl (approved):
Looks good. Thanks for this. Let's get CI sorted out

JosiahBull:
@ahl Resolved CI issues. 😄

Issue #994, opened 2024-12-09
Fails to parse Open Subsonic API


KamWithK:
I was looking at the [Open Subsonic client](https://audioling.github.io/open-subsonic-api-client/) and wanted to use it in Rust, but I'm getting errors:
```
   Compiling funkyduck v0.1.0 (/home/kamwithk/repos/funkyduck)
error[E0432]: unresolved import `super`
  --> src/main.rs:35:5
   |
35 |     generate_api!("openapi.json");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `Client` in the root
   |
   = note: this error originates in the macro `generate_api` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this struct instead
   |
35 |     reqwest::Client;
   |     ~~~~~~~~~~~~~~~

error[E0599]: the method `to_string` exists for reference `&Vec<String>`, but its trait bounds were not satisfied
  --> src/main.rs:35:5
   |
35 |     generate_api!("openapi.json");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method cannot be called on `&Vec<String>` due to unsatisfied trait bounds
   |
   = note: the following trait bounds were not satisfied:
           `Vec<std::string::String>: std::fmt::Display`
           which is required by `Vec<std::string::String>: ToString`
           `&Vec<std::string::String>: std::fmt::Display`
           which is required by `&Vec<std::string::String>: ToString`
           `[std::string::String]: std::fmt::Display`
           which is required by `[std::string::String]: ToString`
   = note: this error originates in the macro `generate_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: the method `to_string` exists for reference `&&Vec<f64>`, but its trait bounds were not satisfied
  --> src/main.rs:35:5
   |
35 |     generate_api!("openapi.json");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method cannot be called on `&&Vec<f64>` due to unsatisfied trait bounds
   |
   = note: the following trait bounds were not satisfied:
           `&Vec<f64>: std::fmt::Display`
           which is required by `&Vec<f64>: ToString`
           `&&Vec<f64>: std::fmt::Display`
           which is required by `&&Vec<f64>: ToString`
           `Vec<f64>: std::fmt::Display`
           which is required by `Vec<f64>: ToString`
           `[f64]: std::fmt::Display`
[...]

Issue #409, opened 2023-04-04
unclear requirement on query types, compilation of compiled code errors out


drahnr:
I am trying to generate some German tax related foo and an API service, but both with the `cli` program and from the builder pattern I get the following (after replacing the `chrono::*` types with custom impls, since they complain about lack of `to_string` impls too):

```
error[E0599]: the method `to_string` exists for reference `&ExportInvoiceSevQuery`, but its trait bounds were not satisfied
     --> src/codegen.rs:16773:43
      |
1159  |     pub struct ExportInvoiceSevQuery {
      |     --------------------------------
      |     |
      |     doesn't satisfy `ExportInvoiceSevQuery: ToString`
      |     doesn't satisfy `ExportInvoiceSevQuery: std::fmt::Display`
...
16773 |         query.push(("sevQuery", sev_query.to_string()));
      |                                           ^^^^^^^^^ method cannot be called on `&ExportInvoiceSevQuery` due to unsatisfied trait bounds
      |
      = note: the following trait bounds were not satisfied:
              `ExportInvoiceSevQuery: std::fmt::Display`
              which is required by `ExportInvoiceSevQuery: ToString`
              `&ExportInvoiceSevQuery: std::fmt::Display`
              which is required by `&ExportInvoiceSevQuery: ToString`
note: the following trait must be implemented
     --> /home/bernhard/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:751:1
      |
751   | pub trait Display {
      | ^^^^^^^^^^^^^^^^^
```

I am not entirely sure what the expectation is, am I supposed to provide the `to_string` impl i.e. with a custom trait in local scope or is this a generation bug?

Any help would be appreciated.

ahl:
Are you able to share the OpenAPI document?

drahnr:
I just invited you to the repo in question, I don't want to publish anything taxation related (which it is). All the hacks are included there.

ahl:
Progenitor doesn't properly handle `query` type parameters that have a non-scalar schema. After reading the spec some more, I'm still not quite clear on how a nested object should appear in the query parameters.

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#parameter-object

[...]

PR #1017, opened 2024-12-23
handle array-like and map-like query parameters

### LINKED ISSUE #268: Support for array / object parameters
We currently attempt to call `.to_string` on query argument values when constructing requests, but this fails on structures that don't implement display (i.e. Vec):
```json
{
  "in": "query",
  "name": "include",
  "schema": {
    "type": "array",
    "items": {
      "type": "string",
      "enum": [
        "custom_fields"
      ]
    }
  }
}
```
This results in a generated client that fails to compile.

The `query` method on `reqwest::RequestBuilder` accepts `T: Serialize`, but due to using `serde urlencoded` internally it isn't able to support `Vec<T>` by default either. The reasoning behind is a lack of specification for how they should be handled (https://github.com/seanmonstar/reqwest/issues/1079#issuecomment-741206109, https://github.com/nox/serde_urlencoded/issues/52).

The Open API spec does allow for array types though in query parameters (https://spec.openapis.org/oas/v3.0.3#parameter-locations). The serialization format is defined via the `style` and `explode` fields.

### LINKED ISSUE #692: Query param of `type: array` with `items.type: string` leads to `to_string()` and unsatisfied trait on `&&Vec<String>`

ahl:
Closes #268, #409, #692, #724, #994.

Query parameters can be scalar values (string, numbers, etc.) today, but per the spec they may be arrays or objects. This PR addresses that gap by adding a proxy `serde::Serializer` (called `progenitor_client::QueryParam`) that adds proper handling of arrays. It also modifies code generation to properly handle all valid input types for query parameters. See #692 for a discussion of the alternatives.

Here's an example of the before and after for generated code:

before
```rust
            let mut query = Vec::with_capacity(1usize);
            if let Some(v) = &minseq {
                query.push(("minseq", v.to_string()));
            }
            #[allow(unused_mut)]
            let mut request = client
                .client
                .get(url)
                .header(
                    reqwest::header::ACCEPT,
                    reqwest::header::HeaderValue::from_static("application/json"),
                )
                .query(&query)
                .build()?;
```

after
```rust
            #[allow(unused_mut)]
            let mut request = client
                .client
                .get(url)
                .header(
                    reqwest::header::ACCEPT,
                    reqwest::header::HeaderValue::from_static("application/json"),
                )
                .query(&progenitor_client::QueryParam::new("minseq", &minseq))
                .build()?;
```

Note that due to behavior in `reqwest` described in #692, this does not address the existing gap regarding query parameters that are not `style=form` and `explode=true` (i.e. the default).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,11 +41,9 @@ jobs:
       with:
         components: rustfmt
     - name: Install stable
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@stable
       with:
-        toolchain: stable
         components: rustfmt
-        default: true
     - name: Build
       run: cargo build --locked --tests --verbose
     - name: Run tests
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -420,7 +420,7 @@ checksum = "97369cbbc041bc366949bc74d34658d6cda5621039731c6310521892a3a20ae0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.90",
+ "syn 2.0.91",
 ]
 
 [[package]]
@@ -1088,7 +1088,7 @@ checksum = "1ec89e9337638ecdc08744df490b221a7399bf8d164eb52a665454e60e075ad6"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.90",
+ "syn 2.0.91",
 ]
 
 [[package]]
@@ -2179,7 +2179,7 @@ checksum = "c8af7666ab7b6390ab78131fb5b0fce11d6b7a6951602017c35fa82800708971"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.90",
+ "syn 2.0.91",
 ]
 
 [[package]]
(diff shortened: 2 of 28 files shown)


ahl:
@MichielHegemans my apologies for not reviewing #724, but I would greatly appreciate your review of this PR.

ahl:
Reviews still welcome, but I'm going to merge this

PR #1079, opened 2025-03-03
Add option to select HTTP client type

wfchandler:
Currently `reqwest::Client` is the only type we allow as the underlying HTTP client. Some consumers of Progenitor may want to use `reqwest_middleware` to enable conveniences like automatic retries and support for `tracing`.

Add a new `with_client_type` argument for `GenerationSettings` to allow callers to select which of these types to use as the client.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1594,6 +1594,7 @@ dependencies = [
  "rand",
  "regress",
  "reqwest",
+ "reqwest-middleware",
  "schemars",
  "serde",
  "serde_json",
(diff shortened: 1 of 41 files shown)


wfchandler:
I think I can get this working with pre and post hooks, closing this out.

eflukx:
Supporting ReqwestClientWithMiddleware would make sense if you'd ask me.

I'm trying to implement a client for an API that requires OAuth2 token requests and renewals. Currently having a hard time doing so as progenitor requires the `reqwest::Client` type. Currently trying to implement this using hooks. I personally prefer to use a pre-generated client over the macro generation, but afaik that doesn't support hooks. Using reqwest middleware would be _the_ nicest solution to this IMO.

PR #1100, opened 2025-03-22
doc: Add note on how to use builder style in build.rs and how to injec…

Tuurlijk:
Add note on how to use builder style in build.rs and how to inject default headers into the client

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ Progenitor can generate two distinct interface styles: positional and builder
 (described below). The choice is simply a matter of preference that many vary
 by API and taste.
 
-## Positional (current default)
+### Positional (current default)
 
 The "positional" style generates `Client` methods that accept parameters in
 order, for example:
@@ -252,7 +252,7 @@ let result = client.instance_create(org, proj, body).await?;
 Note that the type of each parameter must match precisely--no conversion is
 done implicitly.
 
-## Builder
+### Builder
 
 The "builder" style generates `Client` methods that produce a builder struct.
 API parameters are applied to that builder, and then the builder is executed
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -347,3 +347,54 @@ let result = client
...
+    let mut out_file = std::path::Path::new(&std::env::var("OUT_DIR").unwrap()).to_path_buf();
+    out_file.push("codegen.rs");
+
+    std::fs::write(out_file, content).unwrap();
+}
+```
+
+## Changing default client settings
+
+Currently, the generated code doesn't deal with request headers. To add default headers to all requests, you can use the default_headers method when constructing the Client.
```
ahl, line 377: Does it not? This is useful regardless, but I thought we accepted header parameters.

README.md
```diff
@@ -347,3 +347,54 @@ let result = client
...
+
+    let client_with_custom_defaults = reqwest::ClientBuilder::new()
+        .connect_timeout(Duration::from_secs(15))
+        .timeout(Duration::from_secs(15))
+        .default_headers(headers)
+        .build()
+        .unwrap();
+
+    let client = Client::new_with_client(baseurl, client_with_custom_defaults);
+
```
ahl, line 399:
```suggestion
```

README.md
```diff
@@ -347,3 +347,54 @@ let result = client
...
+```
+
+## Changing default client settings
+
+Currently, the generated code doesn't deal with request headers. To add default headers to all requests, you can use the default_headers method when constructing the Client.
+
+```rust
+    let baseurl = "https://company.com/api/v2";
+    
+    let access_token = "OP7nhQY46jf3I2sqyTZBHWEOIRsldfWEFlkjRkrjlER";
```
ahl, line 382: would it make sense to have an example that reads an environment variable to encourage people not to accidentally leak credentials?

ahl (changes requested):
This is awesome and very much appreciated. One question and two suggestions for you. Thank you for the contribution!

README.md
Tuurlijk: I don't see any header autocomplete options pop up when using the Builder Interfacestyle generated api.

ahl (approved):

README.md
```diff
@@ -347,3 +347,53 @@ let result = client
...
+```
+
+## Changing default client settings
+
+Currently, the generated code doesn't deal with request headers. To add default headers to all requests, you can use the default_headers method when constructing the Client.
+
+```rust
+    let baseurl = env!("API_URL").to_string();
+    
+    let access_token = env!("API_ACCESS_TOKEN").to_string();
```
ahl, line 382: I think that `env!` captures the value of the environment variable at compile time. I'd suggest
```suggestion
    let baseurl = std::env::var("API_URL").expect("$API_URL not set");
    
    let access_token = std::env::var("API_ACCESS_TOKEN").expect("$API_ACCESS_TOKEN not set);
```
... or something like that.

ahl (changes requested):

README.md
Tuurlijk: I need to seat a bearer token for each request.

ahl (approved):
thanks again

PR #1116, opened 2025-04-13
Implement From<Infallible> for client error

FliegendeWurst:
This is required if the spec contains schema types like

```
"CustomId": {
  "type": "string",
  "example": "WUjhaGp4EKah_ur11rSfHkzeV",
},
```

### DIFF
--- a/progenitor-client/src/progenitor_client.rs
+++ b/progenitor-client/src/progenitor_client.rs
@@ -295,8 +295,8 @@ impl<E> Error<E> {
 }
 
 impl<E> From<std::convert::Infallible> for Error<E> {
-    fn from(_: std::convert::Infallible) -> Self {
-        unreachable!()
+    fn from(x: std::convert::Infallible) -> Self {
+        match x {}
     }
 }
 
--- a/progenitor-client/src/progenitor_client.rs
+++ b/progenitor-client/src/progenitor_client.rs
@@ -294,6 +294,12 @@ impl<E> Error<E> {
     }
 }
 
+impl<E> From<std::convert::Infallible> for Error<E> {
+    fn from(_: std::convert::Infallible) -> Self {
+        unreachable!()
+    }
+}
+
 impl<E> From<reqwest::Error> for Error<E> {
     fn from(e: reqwest::Error) -> Self {
         Self::CommunicationError(e)


ahl:
Can you explain the necessity for this? What fails or is unusable without this? Thanks!

FliegendeWurst:
Here is the full OpenAPI spec I'm trying to generate a wrapper for: https://gist.github.com/FliegendeWurst/03fa7c5afd778557804a14474e5a7a50

The relevant part (BranchId in components/schemas) I already put in the PR description.

The reason for adding this impl is that the generated code will have a conversion function from string to BranchId(string), with an Error type of Infallible. So the common error type must implement From for Infallible.

ahl:
Thanks for this. I generated a client from that spec and didn't see a build error. Can you share more about the error you're seeing?

FliegendeWurst:
Oh, it seems the error is already fixed in git master. In that case a new patch release would be nice :)

For progenitor `0.9.1` I got:

```
error[E0277]: `?` couldn't convert the error to `progenitor::progenitor_client::Error`
 --> src/lib.rs:2:29
  |
2 | generate_api!("openapi.json");
  |                             ^ the trait `From<Infallible>` is not implemented for `progenitor::progenitor_client::Error`
  |
  = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
  = help: the following other types implement trait `From<T>`:
            `progenitor::progenitor_client::Error<E>` implements `From<InvalidHeaderValue>`
            `progenitor::progenitor_client::Error<E>` implements `From<reqwest::Error>`
  = note: required for `Result<ResponseValue<()>, progenitor::progenitor_client::Error>` to implement `FromResidual<Result<Infallible, Infallible>>`
  = note: this error originates in the macro `generate_api` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
```

FliegendeWurst:
I could also create a more minimal regression test if you're interested.

ahl:
Thanks for that. I see that this was inadvertently fixed by #1067. 

I'm not opposed to merging your change; is there a test we could write today that this change would fix?

FliegendeWurst:
It depends. The generated code would need to attempt to convert an `Infallible` into `progenitor::progenitor_client::Error`. I'm just getting started with progenitor so I don't know if that is still possible with some other API construct.

This impl could also be useful for downstream users that want to reuse the `progenitor_client::Error` directly, not sure if that is common.

ahl (approved):

PR #1131, opened 2025-05-02
Add SDK and CLI compilation test

augustuswm:
This PR extracts a test I was working on in a separate branch. The goal is to verify that the SDKs and CLIs that are generated via tests actually compile.

This is done via a small macro that takes a schema file and outputs the generate SDK and CLI tokens. These tokens are contained within a test function which will fail to compile if the SDK or CLI fail to compile. If they do successfully compile then the test trivially passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1611,6 +1611,10 @@ dependencies = [
 name = "progenitor-impl"
 version = "0.10.0"
 dependencies = [
+ "anyhow",
+ "base64",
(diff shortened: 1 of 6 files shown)


augustuswm:
Secondary issue (`clap::builder::Str: From<std::string::String>`) was due to a missing feature flag in clap. The `string` features was required to be able to construct `PossibleValuesParser`s from Strings.

PR #1140, opened 2025-05-23
update httpmock generation for header parameters

ahl:
... also fix use of the wrong parameter name for query parameters

### DIFF
--- a/progenitor-impl/src/httpmock.rs
+++ b/progenitor-impl/src/httpmock.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 Oxide Computer Company
+// Copyright 2025 Oxide Computer Company
 
 //! Generation of mocking extensions for `httpmock`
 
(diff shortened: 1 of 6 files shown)


progenitor-impl/src/httpmock.rs
```diff
@@ -173,79 +173,136 @@ impl Generator {
...
-                            rr.lifetime =
-                                Some(syn::Lifetime::new("'a", proc_macro2::Span::call_site()));
-                            (Some(quote! { 'a, }), rr.to_token_stream())
-                        } else {
-                            (None, arg_type_name)
-                        };
+                    OperationParameterKind::Query(true) => (
+                        true,
+                        quote! {
+                            Self(self.0.query_param(#api_name, value.to_string()))
```
ahl, line 192: note this was previously using the sanitized name rather than the name from the API (i.e. it was using the wrong name)

PR #1194, opened 2025-09-15
Add an operationId strategy to reject or ignore missing operations

arlyon:
This simply adds a flag to ignore operations with missing ids rather than rejecting them outright. Routes for which progenitor cannot determine a name can be ignored. The default behaviour is not affected. I added a snapshot to exercise the codepath.


arlyon:
This branch was victim to polinrider. While the branch is recovered to its un-compromised state, I am closing this PR for safety reasons.

PR #1195, opened 2025-09-16
Enable custom trait bounds for CLI

wfchandler:
Allow users to specify additional trait bounds on `CliConfig` methods.

### DIFF
--- a/progenitor-impl/src/cli.rs
+++ b/progenitor-impl/src/cli.rs
@@ -5,7 +5,7 @@ use std::collections::BTreeMap;
 use heck::ToKebabCase;
 use openapiv3::OpenAPI;
 use proc_macro2::TokenStream;
-use quote::{format_ident, quote};
+use quote::{format_ident, quote, ToTokens};
 use typify::{Type, TypeEnumVariant, TypeSpaceImpl, TypeStructPropInfo};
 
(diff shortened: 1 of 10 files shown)


wfchandler:
> I'm not a huge fan of this--the whole point of including `JsonSchema` as a trait bound was to satisfy this use case. If it's not useful then perhaps we remove it.

@ahl removing the `JsonSchema` bound would be a breaking change for some of our projects that generate a CLI[^1][^2][^3]. `JsonSchema` is clearly useful in some scenarios, but I think the type reflection in https://github.com/oxidecomputer/oxide.rs/pull/1221 is simpler with the proc macro approach.

[^1]: (https://github.com/oxidecomputer/four-star/blob/7c6938cedde91ef06172a1218ba244a07556fe73/four-star-cli/src/main.rs#L351-L357)
[^2]: (https://github.com/oxidecomputer/rfd-api/blob/ff9dd5e953779f9fdf107d90b5226bcbbc896ad1/rfd-cli/src/main.rs#L378-L381)
[^3]: (https://github.com/oxidecomputer/mfg-support/blob/aa9fb432f549f6aea9eb954877d0c019b744ccf5/mfg-support-cli/src/main.rs#L381-L386).

PR #1210, opened 2025-10-03
remove some dead code

jmpesp:


ahl (approved):
Interesting; did you happen to see when this became dead?

jmpesp:
> Interesting; did you happen to see when this became dead?

I didn't no, it just came up when I did `cargo test`

PR #1261, opened 2025-12-19
fix: support httpmock generation for array types

### LINKED ISSUE #1260: httpmock generation fails for query params with array types (e.g., Vec<T>)
The httpmock generator currently stringifies all query/header/path params via `value.to_string()` (see progenitor-impl/src/httpmock.rs). This implicitly requires Display and only supports single-value params.

For array query params (e.g., Vec<T>), Display isn’t implemented and the generated mocks won’t compile.

It also doesn’t match the client’s actual query encoding (QueryParam uses serde and supports arrays by repeating the key). Should switch httpmock query matching to use the same serialization logic as the client.

danieleades:
closes #1260

Http mock generation fails for me at compile time because it assumes that all types implement Display. This doesn't hold for vectors.

 ## Summary

  - Fix httpmock query-parameter generation for non-`Display` types by reusing `QueryParam` serialization and matching on decoded `(key, value)` pairs.
  - Add `query_param_pairs` in `progenitor-client` to mirror `QueryParam` rules (scalars, arrays with repeated keys, objects/maps), and use it throughout generated httpmock matchers.
  - Add compile-only coverage plus a minimal OpenAPI spec for array query parameters.

 ## Notes
  - The matcher helper applies decoded pairs, which is why this shows up in many places (not just arrays).
  - Serialization failures now `expect` because these are non-recoverable errors in generated mocks/tests.


danieleades:
this now works against my openAPI spec.

note the need to convert &str to String to satisfy the trait bounds of httpmock `json_body_obj`. That bound appears however to be a bug- have opened a PR upstream - https://github.com/httpmock/httpmock/pull/206

danieleades:
@ahl - would it be possible to trigger CI for this one?

this issue is a blocker for me adopting `progenitor`

PR #1267, opened 2025-12-27
chore: configure and check the MSRV

danieleades:


danieleades:
@ahl is this (and my other PRs) of interest?

PR #1349, opened 2026-04-13
Accept `--json-body -` to read body from stdin.

jmcarp:
So that we can write something like `echo '{...}' | oxide instance create --json-body -`. Analogous to the existing `oxide api --input -`.


ahl:
Will take a look; `--json-body /dev/stdin` may be a workaround in the meantime.

ahl (commented):
This adds kind of a truckload of code for each operation; what do you think about having a `progenitor-cli` crate or some additions to the client crate to avoid repeating this code so much?

PR #1356, opened 2026-04-28
Export replaced types

inickles:
This exports the types that someone includes in `replace` when using the `generatie_api!` macro, making them available to users.

### DIFF
--- a/progenitor-impl/src/lib.rs
+++ b/progenitor-impl/src/lib.rs
@@ -374,6 +374,22 @@ impl Generator {
 
         let types = self.type_space.to_stream();
 
+        // Re-export replaced types from the types module so callers can access
+        // them without knowing the replacement path.
+        let mut sorted_replace: Vec<_> = self.settings.replace.iter().collect();
+        sorted_replace.sort_by_key(|(name, _)| name.as_str());
+        let replaced_type_reexports =
+            sorted_replace.iter().map(|(type_name, (replace_path, _))| {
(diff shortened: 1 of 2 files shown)


progenitor-impl/src/lib.rs
```diff
@@ -374,6 +374,22 @@ impl Generator {
...
+            sorted_replace.iter().map(|(type_name, (replace_path, _))| {
+                let type_ident = quote::format_ident!("{}", type_name);
+                let replace_tokens: TokenStream =
+                    replace_path.parse().expect("invalid replacement type path");
+                quote! {
+                    #[allow(unused_imports)]
+                    pub use #replace_tokens as #type_ident;
+                }
+            });
+        let replaced_type_reexports = quote! { #(#replaced_type_reexports)* };
```
ahl, line 391: as discussed, I think this may make more sense in typify.

Do we think this should be the only behavior or should be opt-in (or opt-out)?

What if there's a replace imperative that's never used? Is that caught elsewhere? Would we want to check for that?

Would we want to do something for the other ways to replace a type such as `x-rust-type`.

ahl (approved):
I think it would be helpful to understand the utility of this change, and I'd like to see how this is materialized in the generated code.

Also the style of the code here doesn't particularly match the code around it.

ahl (changes requested):
I mean to say: let's change this (not approved)

inickles:
Closing in place of https://github.com/oxidecomputer/typify/pull/1007

PR #1363, opened 2026-05-15
Stop activating schemars chrono/uuid1 features in progenitor

### LINKED ISSUE #1362: Allow selecting date/date-time backend
[`chrono`](https://crates.io/crates/chrono) is nice, but the ecosystem also has [`time`](https://crates.io/crates/time) and [`jiff`](https://crates.io/crates/jiff) to offer. At the time of writing, `chrono` has 557M downloads, `time` beating that with 665M, and `jiff` trailing behind with 106M. Each of these is clearly used often enough though, that there's going to be people who want to use a specific one of these three to match the rest of their stack.

I'd be willing to implement this myself, but before I spend the time, is this something that you're interested in?

jcgruenhage:
These impls are only needed when consumers derive `schemars::JsonSchema` on generated types that contain `date` or `uuid` fields -- a consumer concern, not progenitor's. `example-macro` and the `README` now show the pattern.

Resolves #1362

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,16 @@ generate_api!(
 Note that the macro will be re-evaluated when the `spec` OpenAPI document
 changes (when its mtime is updated).
 
+If you derive `schemars::JsonSchema` on generated types (see `derives = [ ...
+]` in the macro example above), depend on `schemars` yourself and enable the
+matching feature for each formatted type that appears in your spec, so the
+`JsonSchema` impls for those types are available:
+
+```toml
+[dependencies]
+schemars = { version = "0.8", features = ["chrono", "uuid1"] }
+```
+
 ### `build.rs`
 
 Progenitor includes an interface appropriate for use in a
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -91,6 +91,16 @@ generate_api!(
 Note that the macro will be re-evaluated when the `spec` OpenAPI document
 changes (when its mtime is updated).
 
+If you derive `schemars::JsonSchema` on generated types (see `derives = [ ...
+]` in the macro example above), depend on `schemars` yourself and enable the
+matching feature for each formatted type that appears in your spec, so the
+`JsonSchema` impls for those types are available:
```
ahl:
```suggestion
If you derive `schemars::JsonSchema` on generated types (see `derives = [ ...
]` in the macro example above), add a dependency on `schemars` and enable the
necessary features for each formatted type that appears in your spec. For example:
```

ahl (approved):

ahl:
Thanks for this.

PR #1380, opened 2026-06-29
CLI gen: only define `about`, not `long_about`

david-crespo:
Explanation in code comment. I was dismayed to learn that in our CLI, `-h` and `--help` display different information. clap does this when you define both `about` and `long_about`. This PR stops doing that.

A less invasive alternative would be to keep defining `about` and `long_about`, but include both `summary` and `description` in `long_about` instead of just `description`. That would still be an improvement on the status quo, so I'd be in favor of it, but it does not fix the fact that `-h` and `--help` do different things, which to me is a little bizarre.

### Before

`-h` displays the short description, `--help` displays the long one, but they are disjoint.

image

image

### After

image

The downside is we lose the cute listing of one-liners because the full descriptions get jammed in there. Personally I prefer this overall and I'm not sure more info in the list is necessarily a bad thing, but I can see how someone might disagree.

image

### DIFF
--- a/progenitor-impl/src/cli.rs
+++ b/progenitor-impl/src/cli.rs
@@ -190,15 +190,31 @@ impl Generator {
             consumer: consumer_args,
         } = self.cli_method_args(method);
 
-        let about = method.summary.as_ref().map(|summary| {
+        // A clap command has two help fields, `about` and `long_about`. When
+        // both are set, `-h` (and the parent's subcommand listing) shows
+        // `about`, while `--help` shows `long_about`. The natural mapping --
+        // short summary into `about`, description into `long_about` -- exploits
+        // that to keep listings to one line, but it means `-h` and `--help`
+        // behave differently, and `-h` never shows the description, and
+        // `--help` is missing the summary.
+        //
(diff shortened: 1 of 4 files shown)


david-crespo:
Hand-written commands in the CLI will still have the `about`/`long_about` split, but in the better way that I describe above as the less-invasive alternative.

### 🤖 how hand-written commands in oxide.rs work

Every hand-written subcommand (`cmd_api`, `cmd_auth`, `cmd_disk`, `cmd_instance`, `cmd_net`, `cmd_support_bundle`, `cmd_completion`) is a `#[derive(Parser)]` with `#[command(verbatim_doc_comment)]` and a multi-paragraph `///` doc comment. clap's derive macro maps that the *same way* progenitor used to:
 
- **first paragraph** (up to the first blank line) → `about`
- **whole comment** → `long_about`
 
So e.g. `oxide api -h` shows only:

> Make an authenticated HTTP request to the Oxide API and print the response.

while `oxide api --help` shows the full multi-paragraph prose about `--field`, `@file`, `--paginate`, etc.

ahl (approved):

Issue #843, opened 2024-06-25
Both normalize_doc_attributes and wrap_comments require an unstable toolchain


planckdz:
The formatting options used by the cargo install version (https://github.com/oxidecomputer/progenitor/blob/main/cargo-progenitor/src/main.rs#L94) use formatting options that are documented as unstable:

* https://rust-lang.github.io/rustfmt/?version=v1.7.1&search=#normalize_comments
* https://rust-lang.github.io/rustfmt/?version=v1.7.1&search=#wrap_comments

```
RUST_BACKTRACE=1 cargo progenitor -i api-definition.yml -o output -n output -v 0.1.0
-----------------------------------------------------
 TYPE SPACE
-----------------------------------------------------
....
-----------------------------------------------------

thread 'main' panicked at /Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-progenitor-0.7.0/src/main.rs:101:68:
called `Result::unwrap()` on an `Err` value: Unstable("wrap_comments, normalize_doc_attributes")
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: cargo_progenitor::reformat_code
   4: cargo_progenitor::main
```

ahl:
Yeah: I'm pretty unhappy with the choice I made here... and with the options available. I'd like to use `rustfmt` for formatting so that the generated code could pastt `cargo fmt`... but it's not feasible to use it as a library and shelling out has many problems--including the one you've noted. At this point, I'm inclined to switch everything over to `prettyplease` and see if that's a better experience. PRs welcome or I'll try to get to it in the next week or two.

rgreinho:
Is there a way to disable these formatting options?

maciek134:
FYI didn't work for me in a project already using nightly through `rust-toolchain.toml` - it kept trying to run stable rustfmt until I forced it with `RUSTFMT=$HOME/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rustfmt`.

To be fair, since this is generating visible code maybe it doesn't make sense to force any specific format outside just having something generally readable? At least in my case I'll reformat it with my own rustfmt config anyway

PR #1402, opened 2026-07-25
Switch to prettyplease formater

rpolishchuk:
Not everyone is living on the edge (see https://github.com/oxidecomputer/progenitor/issues/843) :)

This commit switches to `prettyplease` (noted as an option in https://github.com/oxidecomputer/progenitor/issues/843#issuecomment-2189414102) fix test failures with the stable toolchain similar to: 

```
called `Result::unwrap()` on an `Err` value: Unstable("wrap_comments, normalize_doc_attributes, format_strings")
```

Also `split_multiline_doc_attributes` preprocess comments to avoid switch from `///` to `/** */`


ahl:
I'm not sure that this is the right approach for `cargo-progenitor` in particular. It emits a stand-alone crate--I think there's a reasonable expectation for that crate to be `rustfmt` compliant. I do love `prettyplease`, but its output isn't the same (as evidenced by the 45k lines of change in this PR.

ahl:
Happy to discuss this, but if I don't hear from you, I'll close this next week; we can always re-open

